PR #8, opened 2022-04-21
Add storage high level diagram

glimchb:


dandaly (commented):
Hi,
  Some comments on the Device side.  NVMe is a controller abstraction, while virtio-blk is a block device abstraction.  I think you would want to have both, how do you propose that we align this?  Right now in the IPDK object model I have a 'disk' abstraction that below it is structured as having a configuration and statistics object.  I think this covers the space for 'disks', but doesn't cover some of the things you can do in NVMe like namespaces.  Should we start with the disk and expand from there, or start from the controller and make certain things non-configurable for a 'virtio-blk controller'?

Where we are trending here is to have two types of virtual devices for storage:  a disk and a controller.  An NVMe virtual device comes up as a controller, that can have multiple disks inside (one per namespace).  A virtio-blk device comes up just as a disk, so have the configuration object model mirror the PCIe object model space.

glimchb:
Hey,
I agree. virtio-blk and nvme have some similar and some different APIs. I was also thinking how can we abstract this.
 
The proposed diagram is just a visualization draft to start a conversation. Many elements are still missing.
For example, the whole storage services section is missing, i.e.
- compression APIs
- encryption APIs
- raid APIs
- multipath APIs
- lvm APIs
- ...

sandersms (approved):
LGTM

PR #19, opened 2022-06-09
Add NVMe to NVMeOF storage offload API

sburla-marvell:
This is a C based API to dynamically provision an NVMe local to NVMeOF
capable DPU to add/remove subsytems, add/remove remote namespaces and
hotplug/unplug the controllers to host, query stats etc.
We are also working on JSON RPC version of this API.

Signed-off-by: Satananda Burla <sburla@marvell.com>

### DIFF
--- a/storage/nvm_nvmof_offload_api.h
+++ b/storage/nvm_nvmof_offload_api.h
@@ -1,0 +1,455 @@
+/*
+ * Copyright (C) 2022 Marvell International Ltd.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+* @file nvm_nvmof_offload_api.h
+
+* @brief this header file provides the API to interact
+* with an NVMe to NVMeOF storage offload capable subsytem to
+* query capabilities, create/remove local NVM subystems
+* associate/disassociate local controllers exposed to Host
+* with remote namespaces. 
+*
+* @author Satananda Burla
+*/
+
+#ifndef __NVM_NVMOF_OFFLOAD_API_H__
+#define __NVM_NVMOF_OFFLOAD_API_H__
+
+#define NVM_SUBNQN_SIZE 256
+#define NVM_MN_SIZE 40
+#define NVM_SN_SIZE 20
+#define NVM_NGUID_SIZE 16
+#define NVM_EUI64_SIZE 8
+#define NVM_UUID_SIZE 16
+
+struct nvm_offload_cap {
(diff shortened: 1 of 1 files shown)


glimchb:
@sburla-marvell we need to look into https://github.com/spdk/spdk/blob/master/doc/sma.md
there are already proto definitions for that in place https://github.com/spdk/spdk/tree/master/python/spdk/sma/proto

can you identify the gaps between what you posted and SMA ?

sburla-marvell:
> @sburla-marvell we need to look into https://github.com/spdk/spdk/blob/master/doc/sma.md there are already proto definitions for that in place https://github.com/spdk/spdk/tree/master/python/spdk/sma/proto
> 
> can you identify the gaps between what you posted and SMA ?

Yes. we looked at SMA,  There are some missing things in SMA as it seems to be designed for emulation than passthrough, the use case we had was to expose local nvme controllers to host and connect them to remote (nvmeof) namespaces. In sma, the interface to create local devices exposed to host is minimal (create/remove device and attach/detach volume to device), which is good for abstraction but leaves many NVMe specific options missing when emulating NVMe devices (like if I wanted to show a different NQN to host in the NVMe which is different from SUBNQN of the volume attached, nsid values of the volumes etc). I think we would need to add NVMe emulation interface in SPDK, which could then help in creating local NVMe devices with all customization needed(like NQNs, nsids, maintaining each controller's stats etc which could be different from what you get from querying the remote bdev itself ), May be this is already implemented but I have not seen it. Apart from this though abstraction is good, it sometimes leaves out important performance items associated with physical devices , like customers wanted to control number of SQ/CQ pairs going to a NVMe PF/VF, queue sizes allowed etc, which are very much an NVMe specific things but important enough for performance. I dont know if abstracting such options across virtio-blk , nvme, virtio-scsi etc would make sense.  Our customers had many more requirements for  locally emulated nvme controllers (being able to specify different options for SUBNQNs, nsids etc, queue sizes etc) which we were addressing in this API.  Bottom line, this api can live one level below the sma specification as a plugin,  but provides more customization than SMA for NVMe devices.

glimchb:
@sburla-marvell I agree that SMA is missing a lot of stuff we need... I was also thinking is we make this into SMA plugin... or do something else with it... let's keep talking

glimchb (approved):
it's a good starting point, i'm ok with merging and continue discussion on top of that
obviously we need this to be protobuf

PR #28, opened 2022-06-17
storage: enhance spec with nvme pic and stats

glimchb:


sandersms (approved):
Looks good

storage/spec.md
```diff
@@ -81,6 +92,12 @@ It should have all the correct controllable parameters according to NVMe spec.
 |----------------|---------------------|----------------------------------------------|
 | string         | Subsystem NQN       | NVMe subsystem (NQN)                         |
 
+#### `NVMe Subsystem Stats`
+
+| Type           | Parameter           | Details                                      | 
+|----------------|---------------------|----------------------------------------------|
+| string         | Subsystem NQN       | NVMe subsystem (NQN)                         |
+
 #### `NVMe Controller Create`
```
sburla-marvell, line 101: Does controller create also involve exposing the Controller to the host on PCIe bus or is that a separate operation ?

storage/spec.md
```diff
@@ -48,6 +49,14 @@ tbd...
...
 
+![NVMe examplained](nvme-sub-ctrl-ns.png)
+
+3 Objects are identified on this layer:
+- NVMe Subsystem - holding all other objects in NVMe world.
+- NVMe Controller - responsible for Queues and Commands handlings. Have to belong to some subsystem.
+- NVMe Namespace - representing remote namespace. Belongs to a apecific controller (private NS) or shared between controllers (usually for Multipathing).
+
+
 #### `NVMe Subsystem Create`
```
sburla-marvell, line 60: Question on NVMe Subsystem Update. When is this supposed to be used since there is no way in NVMe protocol for for NVMe subsystem update to be propagated to the host ?

storage/spec.md
glimchb: I was thinking it is the same operation, if you see the need, to split, we can discuss, for sure

storage/spec.md
glimchb: I was thinking instead of delete and re-create we can issue update. once driver is re-loaded on the host, changes will appear

sburla-marvell: There is an ambiguity that a subsystem can be associated with multiple controllers. and driver can be loaded/unloaded for each controller separately.  So probably it could be stated as once all controllers in a subsystem are in CSTS.SHN state, then a subsystem update can be issued and any subsequent activation of any controller in the subsystem will have visibility to the update.

storage/spec.md
sburla-marvell: I do not see a need. But wanted to see if others had any thoughts on this.

PR #36, opened 2022-07-14
API gw changes

glimchb:
add bypass linux kernel to HW
add bypass OPI to Vendor SDK


sandersms (approved):
Good with the first set of changes.  Have some additional ideas to work through on the diagram.

PR #41, opened 2022-07-19
Include API Abstraction Layer Detailed Diagram

sandersms:


glimchb:
@sanders-mark unrelated to this, we should move `doc/minutes/images` 1 level up to `doc/images`...

glimchb (commented):
i like this

mestery:
Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.

sandersms:
>

glimchb:
> Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.

isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?

mestery:
> > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> 
> isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?

That's exactly my point: Each vendor exposes their own APIs from the DPU/IPU devices, and this `OPI API Conglomerate` is a clustered application running *off* the DPU/IPU cards which orchestrates them. We're moving quickly into on orchestration play here.

glimchb:
> > > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> > 
> > 
> > isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?
> 
> That's exactly my point: Each vendor exposes their own APIs from the DPU/IPU devices, and this `OPI API Conglomerate` is a clustered application running _off_ the DPU/IPU cards which orchestrates them. We're moving quickly into on orchestration play here.

what if you want to run APP on the DPU and use vendor agnostic shim abstraction layer ?

mestery:
> > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> 
> isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?

Also, in the diagram, vendor APIs are outside the `OPI API Conglomerate`, so they are local, but the conglomerate is remote.

mestery:
> > > > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> > > 
> > > 
> > > isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?
> > 
> > 
> > That's exactly my point: Each vendor exposes their own APIs from the DPU/IPU devices, and this `OPI API Conglomerate` is a clustered application running _off_ the DPU/IPU cards which orchestrates them. We're moving quickly into on orchestration play here.
> 
> what if you want to implement and run new APP on the DPU and use vendor agnostic shim abstraction layer ?

Wouldn't you want to use the OPI APIs to *build* an application across a fleet of DPU/IPU devices? I think both use cases are valid here honestly.

mestery:
> > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> 
> @mestery The intent is to show how the API layer expands. We are not directly calling the vendor APIs since we are requesting as capability such as setting up a VLAN or creating an NVMe storage resource. These actions require a set of vendor and/or standard SDK calls to perform the action. If we were to just call into the vendor APIs, then some higher layer has to track the steps. At the DPU level we would probably want to have those steps contained and provide a response as to the intended action being completed instead of each step in the process being responded to. The protobuf file would contain the config parameters that need to be set; but, each setting may require a different SDK operation.

If the `OPI API Conglomerate` can interface with all the vendor APIs, it becomes a clustered application which runs outside the cards and orchestrates them all. I may be thinking a few steps ahead of where we're at here.

sandersms:
> > > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> > 
> > 
> > isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?
> 
> Also, in the diagram, vendor APIs are outside the `OPI API Conglomerate`, so they are local, but the conglomerate is remote.

Would you suggest that the conglomerate include the Vendor SDKs/APIs?

mestery:
> > > > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> > > 
> > > 
> > > isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?
> > 
> > 
> > Also, in the diagram, vendor APIs are outside the `OPI API Conglomerate`, so they are local, but the conglomerate is remote.
> 
> Would you suggest that the conglomerate include the Vendor SDKs/APIs?

No, those components run on each DPU/IPU card. The conglomerate is orchestrating things across a fleet of cards in what I'm thinking.

sandersms:
> > > > > Looking at this diagram, it makes me wonder why the entire box labeled `OPI API Layer Abstraction` would even need to run on the DPU/IPU cards. Since it's just a shim layer (with a metric crap ton of things wrapped around it), and it's just calling into vendor APIs, it probably makes sense to just allocate a pool of servers to run this thing on.
> > > > 
> > > > 
> > > > isn't vendor APIs local ? libraries... we have to expose over the network those calls... no ?
> > > 
> > > 
> > > Also, in the diagram, vendor APIs are outside the `OPI API Conglomerate`, so they are local, but the conglomerate is remote.
> > 
> > 
> > Would you suggest that the conglomerate include the Vendor SDKs/APIs?
> 
> No, those components run on each DPU/IPU card. The conglomerate is orchestrating things across a fleet of cards in what I'm thinking.

I see it as the conglomerate in the diagram is the API layer construct that runs on the D/IPU.  The client on the outside (local cpu, external entity, etc.) is the orchestrator that is running a fleet of DPUs and will have his own intent mechanism to issue the protobuf messages to the D/IPU.

mkalderon:
@mestery how would you handle the authentication and authorization in that case ? would each service on the xPU have to handle that themselves ? 
And how about one main agent to communicate with to configure, start services etc on the xPU- you'de have to have some level of API conglomerate on the DPU right ?

mkalderon:
@sanders-mark nice illustration - is there some color coding you used here ? why is the OTEL collector outside the Layer Abstraction ? thanks

mestery:
> @mestery how would you handle the authentication and authorization in that case ? would each service on the xPU have to handle that themselves ?
> And how about one main agent to communicate with to configure, start services etc on the xPU- you'de have to have some level of API conglomerate on the DPU right ?

You would definitely need an agent on the DPU/IPU for sure. Now, how that handles authentication and authorization, presumably the vendors have their own way of doing that, which the `OPI API Conglomerate` would make use of when communicating with the cards. If this is not the case, then yes, we'd need to add those pieces to the DPU/IPU.

As we all know, the CPU resources on the cards are limited, and I'm starting to think outside the box a bit in terms of what we want to bring to those CPUs. Rather than replicating the host system on the card CPUs, we should try to optimize things there.

sandersms:
> @sanders-mark nice illustration - is there some color coding you used here ? why is the OTEL collector outside the Layer Abstraction ? thanks

Color coding was simply to delineate some of the aspects to illustrate things.  There are probably some changes to the colors that are still needed to show the alignment better on items.

The OTEL collector is outside of the layer abstraction since it is a collector for many services that may be operating on the xPU.  The API gateway would provide telemetry information to the collector and similarly the network, storage, security, etc. would provide telemetry information from those services.  The collector would then deliver the telemetry data to a service for processing of the information for any actions to be performed.

sandersms:
@mestery @glimchb Updated the diagram based on today's discussion and updated the PR.  I also noted that this view is with the capability on the xPU.  We will need to consider the differences when we think about API Abstraction blob being external to the xPU.

glimchb (approved):
Lgtm

PR #42, opened 2022-07-24
storage: add revisions to SPEC

glimchb:
This is an outline for the SPEC versions as we moove forward


sandersms (approved):
Good addition

PR #43, opened 2022-07-25
Rename Service to Shim API

mestery:


glimchb (approved):
new name is better

sandersms (approved):
Shim naming is fine - will have to update other diagrams also accordingly.

PR #50, opened 2022-07-27
CI: Add golang proto generation

mestery:
This adds simple CI to validate the storage protobuf files on pull
requests and merges to master.

### DIFF
--- a/storage/Makefile
+++ b/storage/Makefile
@@ -1,0 +1,5 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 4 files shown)


storage/Makefile
```diff
@@ -0,0 +1,13 @@
+# Copyright (C) 2022 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
```
glimchb, line 3: I personally do
```
   docker run -v $PWD:/defs namely/protoc-all -d proto -l go -o ./proto/  --go-source-relative
```
but you are the go expert ;-)

glimchb (approved):
YES!

storage/Makefile
mestery: Good idea!

PR #66, opened 2022-08-02
security: Update security APIs

mestery:
Allocate and return the UUID on create requests.


glimchb (approved):
this is good

PR #67, opened 2022-08-02
security: Use UUID instead of int64 foR IDs

mestery:
This creates a UUID message type and uses that for IDs for all objects
rather than an int64.


seroyer (approved):
Looks good

glimchb (approved):
excellent

PR #73, opened 2022-08-03
Updated API GW and Control Path Stacking Diagrams

sandersms:


sandersms:
Updated detailed diagram and new stacking diagram added.

glimchb (approved):
great
I think translation Fn is not part of OPI, but it's ok to commit like this

sandersms:
> great
> I think translation Fn is not part of OPI, but it's ok to commit like this

Agreed.  I need to clarify that the top bubble that covers the REST & gRPC, Fanout, and Protocol Buffer is what OPI is addressing (thus the bubble around the areas).  The translation is the path for the specific function area based on the protocol buffers and provides the interface to the SDK where needed - not part of OPI API.

PR #84, opened 2022-08-09
security: Add name to IPsecCreateRequest message

mestery:
Used when generating the connection via the vici API.


glimchb (approved):
approved

PR #92, opened 2022-08-11
security: dhgroups not dhgroup

mestery:
Commit 4920567 incorrectly added dhcgroup instead of dhgoups to the
Security Associations. This commit fixes that error.


glimchb (approved):
small change, can merge now, no need to wait

PR #94, opened 2022-08-16
security: Revamp security APIs

mestery:
The previous security APIs started from the Ligato IPsec protobuf file
[1]. This was done for expediency. But, it turns out that converting
those into strongSwan data structures is a bunch of work that is easier
to solve by simply redoing the APIs to match the strongSwan data
structures found here [2] and the VICI APIs found here [3].

[1] https://github.com/ligato/cn-infra/blob/master/examples/cryptodata-proto-plugin/ipsec/ipsec.proto
[2] https://docs.strongswan.org/docs/5.9/swanctl/swanctlConf.html
[3] https://www.strongswan.org/apidoc/md_src_libcharon_plugins_vici_README.html


glimchb (approved):
this looks clean and better approach
not sure who prefers Ligato, so can't say for them...

PR #96, opened 2022-08-17
August 11 notes and minutes

sandersms:


glimchb (approved):
suggest to reverse the order of this file so newest meetings are on the top

PR #98, opened 2022-08-24
security: Flesh out IPsecListSas API

mestery:
This updates the return IPsecListSasResp structure which is returned by
the IPsecListSas API call to more closely match the list-sas() vici
command found here [1]. The return structure is assembled in vici here
[2].

[1] https://www.strongswan.org/apidoc/md_src_libcharon_plugins_vici_README.html
[2] https://github.com/strongswan/strongswan/blob/master/src/libcharon/plugins/vici/vici_query.c#L523


security/proto/ipsec.proto
```diff
@@ -390,7 +301,64 @@ message IPsecListSasReq {
 }
 
 message IPsecListSasResp {
```
glimchb, line 303:
```
message IPsecListConnsReq {
	string ike = 1;
}
```
remains the same ?

security/proto/ipsec.proto
```diff
@@ -390,7 +301,64 @@ message IPsecListSasReq {
...
+	string ppk                    = 28;
+	string established            = 29;
+	string rekey_time             = 30;
+	string reauth_time            = 31;
+	repeated string local_vips    = 32;
+	repeated string remote_vips   = 33;
+	repeated string tasks_queued  = 34;
+	repeated string tasks_active  = 35;
+	repeated string tasks_passive = 36;
+	message ListChildSa {
```
glimchb, line 340: assume you only use it here and not in other messages for now...

glimchb (approved):
approved

security/proto/ipsec.proto
mestery: Correct!

security/proto/ipsec.proto
mestery: Yes, because that one is an even bigger change, so wanted to separate it and go one at a time.

PR #100, opened 2022-08-24
security: Fix IPsecListSasResp struct

mestery:
Commit 92694d2 modified the IPsecListSas to have correct return values.
Sadly, I neglected to realize I needed to return a list of IKE_SA
structures here. This commit fixes that.

Also, while here, and as Boris foreshadowed, I moved ListChildSa up a
level and out of what was once the IPSecListSasResp struct, and is now
the ListIkeSa struct.


glimchb (approved):
excellent

PR #101, opened 2022-08-25
Initial Network Config Comparison Table

sandersms:


mestery (changes requested):
This is great! Can you replace spaces in the filename with underscores please? Otherwise, a great start!

pathreya (commented):
This is a good template. Can you please add a row for IPsec? Is Tunnel supposed to Overlay Tunnel? If not an entry for that too would help.

AnuradhaKaruppiah (commented):
Thanks for creating this @sanders-mark. Can I suggest you add another column to this table that conveys how relevant a capability is for xPUs (as what is critical for switches maybe less for for servers/NICs).

sandersms:
> This is great! Can you replace spaces in the filename with underscores please? Otherwise, a great start!

Modified the file name in the amended PR.

sandersms:
> Thanks for creating this @sanders-mark. Can I suggest you add another column to this table that conveys how relevant a capability is for xPUs (as what is critical for switches maybe less for for servers/NICs).

Good Feedback.  This is an initial cut at the table.  We need to continue to iterate on the table to get the columns filled out and add any new ones that are missing.  Agree that the column for the relevance to xPUs is needed - then I can get rid of the note that I started for the lag item as this is more switch related vs NIC.

mestery (approved):
Good first start, we can iterate on this as we go.

doc/Config_API_Matrix.md
```diff
@@ -0,0 +1,29 @@
...
+| Capability     | OpenConfig   | OVS DB | Cyborg | P4 | SONiC | VPP |
+|----------------|--------------|--------|--------|----|-------|-----|
+| Inventory      | oc-platform  |        |        |    |       |     |
+| Interface      | oc-interface |        |        |    |       |     |
+| Ethernet       |              |        |        |    |       |     |
+| VLAN           | oc-vlan      |        |        |    |       |     |
+| 802.1x         |              |        |        |    |       |     |
+| IP             |              |        |        |    |       |     |
+| IPv6           |              |        |        |    |       |     |
+| Tunnel ipv4    |              |        |        |    |       |     |
```
mgheorghe, line 12: GRE, VXLAN ? we should mention each tunneling protocol as separate line items

PR #102, opened 2022-08-29
security: Update README.md

mestery:
This adds more information into the `security/README.md` file detailing
what exactly the OPI Security APIs are currently targeting, and the
proposed architecture around the IPsec portion of the APIs.


security/README.md
```diff
@@ -1,5 +1,18 @@
 # OPI Security APIs
 
+The OPI Security APIs are currently focusing on implementing an IPsec API which
+maps on top of the [strongSwan](https://www.strongswan.org)
+[vici Plugin](https://docs.strongswan.org/docs/5.9/plugins/vici.html). For
+more details on the vici API, you can look at the strongSwan documentation
```
shorman-corigine, line 6: For consistency with vici.html (linked above) I think "vici API" should be "vici protocol" or "vici plugin".

mestery: I agree, will make this change.

glimchb (approved):
this is great now, 
thanks for fixing my small nits

shorman-corigine (commented):
I'm reserving endorsement of the approach proposed here.
But so far as this text documenting the approach goes, I'm happy.
Thanks for addressing my earlier review.

glimchb:
@shorman-corigine we are welcoming the reviews and other opinions... you can comment freely

shorman-corigine:
> @shorman-corigine we are welcoming the reviews and other opinions... you can comment freely

Of course :)
But I need to put my thoughts together first.

glimchb:
I do think we need to separate reference implementation VS generic APIs...
`vici` on this page is just an example... 
everybody else are free to use their own implementation instead of strongswan or vici...
see the picture from https://github.com/opiproject/opi-poc/blob/main/storage/OPI-storage-SPDK-bridge.png

also storage in this repo doesn't talk about specific implementations [storage readme]([../storage/README.md)

PR #103, opened 2022-08-29
security: Remove link to deleted spec

mestery:
Commit 9bd55a5 removed the security spec, but forgot to remove the link
from the README.md file. This fixes that.


glimchb (approved):
thanks for fixing this

PR #105, opened 2022-08-29
security: Add list-[conns,certs] APIs

mestery:
This adds the necessary protobuf glue to make the IPsecListConns and
IPsecListCerts APIs work correctly. The vici protocol defines these
[1] [2] to return slightly different data than the loading of both
connections and certificates.

[1] https://github.com/strongswan/strongswan/blob/master/src/libcharon/plugins/vici/README.md#list-conns
[2] https://github.com/strongswan/strongswan/blob/master/src/libcharon/plugins/vici/README.md#list-certs


glimchb (approved):
reasonable

PR #106, opened 2022-08-29
security: Add missing bits

mestery:
Commit 859ceeb added the fleshed out portion of the IPsecListConns and
IPsecListCerts APIs. However, the strongSwan documentation is not quite
accurate vs. the code. This corrects missing fields which the code can
return but which are not in the documentation.


glimchb (approved):
reasonable

PR #107, opened 2022-08-29
security: Fix IPsecListConnsResp message

mestery:
Commit 859ceeb neglected to change the return value of
IPsecListConnsResp to the newly created ListConnResp message. This
commit fixes that negligence.


glimchb (approved):
thanks for fixing this

PR #113, opened 2022-09-08
OPI xPU Aware CSIs

mkalderon:
Inception. Beginning of document and images trying to define how CSIs will interact with OPI-compatible xPUs in the data center.


glimchb (approved):
option1 makes the most sense to me, assuming DPUs are attached to Hosts... so there is a relationship there... vs Controller, which is one per Storage target system...

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+## High Level Overview of CSI with xPU
+
+Once a xPU is added to a host, the xPU will start handling all the communication to the target and offload all storage processing from host to xPU. The xPU will emulate an nvme device to the host which from  host point of view will look as a local nvme device.
+The xPU needs to receive remote storage information to enable it to connect to the remote target. There are several ways in which this information can be passed to the xPU. In all cases, there will be some OPI gRPC server on the xPU that will be able to receive these configurations in a standard way ( proto-bufs + gRPC )
+
+### Option 1
+
+The first option is that the CSI Node Driver on the host will continue receiving the connection params as it does today and pass them to the xPU. This will require changes only on the Node driver and not the CSI controller.
+
+![CSI without DPUs](../doc/images/CSI-With-xPU-option1.png)
```
seroyer, line 20: This (and the ones below) need different alt text.

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Option 2
+
+The second option is that the CSI Driver Controller will be the one configuring the DPU and the node driver will only be informed of its local nvme device
```
seroyer, line 24: Need to replace the remaining "DPU" terms with "xPU"

glimchb:
@intelfisz fyi

intelfisz:
@glimchb I also assume option 1 to be the right one - the standard flow is associated with controller communicating connection information to node drivers - I see no reason to change that, especially since changes are confined to node driver only in this case.

storage/opi-csi.md
mestery: Agree with Steve.

mestery (changes requested):
The diagrams all have titles with DPUs, can you change this to xPU please?

storage/opi-csi.md
mkalderon: right, will fix

storage/opi-csi.md
mkalderon: Missed this one! will fix

glimchb:
> > option1 makes the most sense to me, assuming DPUs are attached to Hosts... so there is a relationship there... vs Controller, which is one per Storage target system...
> 
> option 3 could reduce the need for making changes per node-driver, will add more layering. But for newer drivers we could provide standard APIs that for their case would avoid the need for OPI node driver and interact directly with OPI controller driver.

we can open this to discussion in the next week API call, I will ask @sanders-mark to add this to the agenda

mestery (approved):
Thanks for the fixes on the naming!

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+## High Level Overview of CSI with xPU
+
+Once a xPU is added to a host, the xPU will start handling all the communication to the target and offload all storage processing from host to xPU. The xPU will emulate an nvme device to the host which from  host point of view will look as a local nvme device.
+The xPU needs to receive remote storage information to enable it to connect to the remote target. There are several ways in which this information can be passed to the xPU. In all cases, there will be some OPI gRPC server on the xPU that will be able to receive these configurations in a standard way ( proto-bufs + gRPC )
+
+### Option 1
+
+The first option is that the CSI Node Driver on the host will continue receiving the connection params as it does today and pass them to the xPU. This will require changes only on the Node driver and not the CSI controller.
+
+![CSI With xPU option1](../doc/images/CSI-With-xPU-option1.png)
```
pohly, line 20: A general comment about the diagrams: it would be useful to describe what is being done in terms of CSI operations (`ControllerPublishVolume`, `NodeStageVolume`, `NodePublishVolume`) because then it is clearer what is meant. As it stands now, for example "during bind/mount to pod" is ambiguous: one typically happens during `NodeStageVolume` (mounting), the other during `NodePublishVolume` (bind to directory for a pod).

pohly: The first option makes sense to me, assuming that there is a communication path between node driver and OPI gRPC. In particular the OPI API call can report back which VF has the intended volume.

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Option 2
+
+The second option is that the CSI Driver Controller will be the one configuring the xPU and the node driver will only be informed of its local nvme device
+
+![CSI with xPU option2](../doc/images/CSI-With-xPU-option2.png)
```
pohly, line 26: So in this case, `ControllerPublishVolume` will ensure that the volume is attached. The `publish_context` must contain information about the VF and will be made available to `NodeStageVolume`. That should work.

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Option 3
+
+The third option which is similar to option 2 is to introduce another layer that is the OPI CSI driver controller. This driver will interact with the different vendor specific CSIs and provide translation between them to the xPU to enable connectivity.
+
+![CSI with xPU option3](../doc/images/CSI-With-xPU-option3.png)
```
pohly, line 32: This looks similar to the idea from https://github.com/intel/oim#storage--csi. The prototype had support for ceph-csi.

The overall question is: who delivers which components and provides support for them?

The third option makes it possible for an xPU vendor to add xPU support to an existing storage solution without the storage vendor doing anything. But then the xPU vendor is responsible for end-customer support.

The other two options invert the responsibility: the storage vendor must do something in their driver for xPUs and thus has the responsibility to validate and support the result.

pohly (commented):
Hello! This looks like and interesting activity and something that is related to my work. I worked on a similar solution a while back (see below) and am one of the upstream developers working on CSI in Kubernetes. I am a SIG Storage reviewer and the maintainer of [csi-test](https://github.com/kubernetes-csi/csi-test).

I missed all other discussions around this, so let me simply dive in and leave some comments on what is in this PR...

storage/opi-csi.md
```diff
@@ -0,0 +1,32 @@
...
+Define OPI interfaces and APIs that will enable vendor CSIs driver developers to utilize xPUs in the data-center with one of the main goals being minimal changes to CSI vendor drivers, and a single API for configuring all OPI-compliant xPUs
+
+## High Level Overview of CSI in non-xPU environment
+
+![CSI without xPUs](../doc/images/CSI-Without-xPU.png)
+
+## High Level Overview of CSI with xPU
+
+Once a xPU is added to a host, the xPU will start handling all the communication to the target and offload all storage processing from host to xPU. The xPU will emulate an nvme device to the host which from  host point of view will look as a local nvme device.
+The xPU needs to receive remote storage information to enable it to connect to the remote target. There are several ways in which this information can be passed to the xPU. In all cases, there will be some OPI gRPC server on the xPU that will be able to receive these configurations in a standard way ( proto-bufs + gRPC )
```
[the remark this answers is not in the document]
glimchb: good point, we haven't dived into secure considerations of the connection between client and server yet

storage/opi-csi.md
glimchb: exactly my though, OPI API call can report back which VF has the intended volume and then CSI node driver can use this information to find correct local NVMe device to bind/mount into the POD

regarding CSI exact operations comment - agree we need more details on this, all diagrams created using https://app.diagrams.net/ so if you have time, you can suggest an edit and help us out here

storage/opi-csi.md
glimchb: technically possible, yes, but I'm worried about scale a bit... in my mind if DPUs and IPUs are plugged into Server, then Node driver should be responsible to configure them... while Controller driver should have minimal changes and still only talk to Storage system... but SME opinion is what we need here, so chime in...

storage/opi-csi.md
pohly: This option could use the CSI controller from the storage vendor and the completely replace the node part. That should work as long as the controller really only handles volume provisioning and never communicates with the node part. At least for ceph-csi that was the case.

pohly:
> @pohly your timing is perfect, we just kicked of this discussion 4 days ago by this pull request and hope to have our `first` meeting this week API meeting https://lists.opiproject.org/g/opi/calendar you are welcome to join

Thursdays around that time are already fairly busy for me.

> also join on slack discussion https://opi-project.slack.com/archives/C03Q50JUWAU

I joined that Slack and will try to follow along.

storage/opi-csi.md
mkalderon: Need to also consider the case where DPU could be on the target side, in which case OPI could provide the means for API between controller + target.

storage/opi-csi.md
mkalderon: wouldn't xPU vendors need to add support for different storage solution anyway? even if they get information from node driver, they would need to know what to do with it -> or would storage providers provide a xPU driver ?

storage/opi-csi.md
glimchb: yes, thank you @mkalderon that is a new interesting use case

storage/opi-csi.md
glimchb: not always, let's say ALL storage products do NVMeoF for block storage (I know this is not the case) then xPU vendor will only need to provide NVMeoF initiator regardless of the target storage system. CSI will come from the storage vendor to provision this target storage system.

in cases where protocol is proprietary either xPU vendor or Storage vendor wil provide xPU driver - this is out of OPI scope anyways

mkalderon: In the case you provided above, all node drivers just need to provide NVMeoF initiator - so how do they differ ? and if they don't differ, then they can be replace entirely by OPI node driver right ?

glimchb: vendors usually provide the bundle Node-Controller as a single binary that can accept argument what personality to run... 
I don't think anybody doing separate deploy of unrelated node and ctrl drivers from different container images... 
let's ask our SME @pohly

pohly: Providing a single binary and thus container image is convenient, but not required. I don't know whether everybody does it.

For the discussion above it doesn't matter: the relevant part is that the vendor has some install instructions for the node part of their driver. Instead of following those, a generic OPI node driver would need to be deployed instead.

PR #116, opened 2022-09-09
security: Add flow offload API for OPI firewall

### LINKED ISSUE #115: Add OPI Security firewall API based on flow-offload
Add a flow-based offload API to integrate with firewalls for offloading flows onto DPUs and IPUs, similar to the api defined in the [sessionOffload API](https://github.com/att/sessionOffload/blob/master/protos/openoffload.proto) of the [sessionOffload project](https://github.com/att/sessionOffload).

mestery:
This commit adds a first cut at an OPI Security API for flow-offload, with the goal being something which firewalls can use to offload flows. This was initially taken from the sessionOffload project [1], specifically the openoffload.proto [2]. I have modified this to make it compile with the latest protobuf compilers, and fixed a few APIs to make a bit more sense as well.

[1] https://github.com/att/sessionOffload
[2] https://github.com/att/sessionOffload/blob/master/protos/openoffload.proto

Closes #115


glimchb (approved):
this looks reasonable
I wonder can we just use/include https://github.com/att/sessionOffload/blob/master/protos/openoffload.proto ? 
why do we need to copy it ?

mestery:
> this looks reasonable
> I wonder can we just use/include https://github.com/att/sessionOffload/blob/master/protos/openoffload.proto ?
> why do we need to copy it ?

The reason is that I had to modify that API, both because it won't compile with the latest protobuf compilers, but also because I wanted to change bits of it.

stefanchulski:
Currently, it is a very narrow use case 5-Tuple flow offload.  Do we want to make it more flexible, similar to TC flower for Kernel netdev interfaces?

glimchb:
sure @stefanchulski, can you propose PR ?

mestery:
> Currently, it is a very narrow use case 5-Tuple flow offload. Do we want to make it more flexible, similar to TC flower for Kernel netdev interfaces?

No one said this is the final API at all, it is a first proposal. It's a lot of work to even get to this. Please feel free to also propose an update to this API if you want. I wanted to get something out there to start hacking on this in the `opi-poc` repository to show how it might work with something as simple as an eBPF/XDP program.

PR #117, opened 2022-09-15
Agenda Topics Update

sandersms:


glimchb (approved):
lgtm

Agenda.md
```diff
@@ -8,6 +8,8 @@ Sept. 15, 2022
 
 Sept 22, 2022
 
+- Marvell demo of Session Offload
```
mestery, line 11: Will be interesting to see this on what I presume is Octeon. I would also like to see how it related to the [OPI Firewall Session Offload API](https://github.com/opiproject/opi-api/blob/main/security/proto/openoffload.proto) I merged. And also, I want to evolve this API in the direction of rte_flow, which seems to make some sense.

mestery (approved):

PR #125, opened 2022-09-21
Networking Interfaces protobuf file

sandersms:
Protobuf file set from the OpenConfig Interfaces model along with the python derived files.


mestery (commented):
This is a nice start @sanders-mark! It looks like your gRPC service exposes listing existing interfaces and updating them. But we don't allow creation of interfaces. Is this because we're only looking at interfaces exposed out of the card (e.g. the physical ports?), or you will update this to allow creation and deletion of virtual interfaces (e.g. what I exposed into the host?).

mestery:
Also, would be good to get some golang generated files here, and add Makefile logic similar to `storage` and `security`.

mestery (changes requested):
I think we should add the Makefile logic and golang generated files before merging. Otherwise, looks good as a first start!

sandersms:
> This is a nice start @sanders-mark! It looks like your gRPC service exposes listing existing interfaces and updating them. But we don't allow creation of interfaces. Is this because we're only looking at interfaces exposed out of the card (e.g. the physical ports?), or you will update this to allow creation and deletion of virtual interfaces (e.g. what I exposed into the host?).

In this case, the OpenConfig doesn't have the virtual interface definition and only the physical ones.  I put in the get, list, and update for the physical ones.  For virtual ones, we probably need to add in the Create to add new instances.

glimchb:
> > This is a nice start @sanders-mark! It looks like your gRPC service exposes listing existing interfaces and updating them. But we don't allow creation of interfaces. Is this because we're only looking at interfaces exposed out of the card (e.g. the physical ports?), or you will update this to allow creation and deletion of virtual interfaces (e.g. what I exposed into the host?).
> 
> In this case, the OpenConfig doesn't have the virtual interface definition and only the physical ones. I put in the get, list, and update for the physical ones. For virtual ones, we probably need to add in the Create to add new instances.

we can always add them later

sandersms:
> this is a good start conside adding makefile like this to network folder https://github.com/opiproject/opi-api/blob/main/storage/Makefile

Yes, this needs to be added; but, I have to work the command for the .py file generation from the grpc_tools.protoc command vs standard protoc.

sandersms:
> Also, would be good to get some golang generated files here, and add Makefile logic similar to `storage` and `security`.

Agreed.  Need to add the golang option for these files

mestery:
> > > This is a nice start @sanders-mark! It looks like your gRPC service exposes listing existing interfaces and updating them. But we don't allow creation of interfaces. Is this because we're only looking at interfaces exposed out of the card (e.g. the physical ports?), or you will update this to allow creation and deletion of virtual interfaces (e.g. what I exposed into the host?).
> > 
> > 
> > In this case, the OpenConfig doesn't have the virtual interface definition and only the physical ones. I put in the get, list, and update for the physical ones. For virtual ones, we probably need to add in the Create to add new instances.
> 
> we can always add them later

Yup, I agree, lets add them afterwards in a subsequent PR.

dandaly:
Hi Mark,
  Is the thought that we would implement in OPI a protobuf server that would unpack these messages and convert them into OpenConfig?  What is the motivation for having this on top of OpenConfig which already has a gNMI RPC layer?

sandersms:
> Hi Mark, Is the thought that we would implement in OPI a protobuf server that would unpack these messages and convert them into OpenConfig? What is the motivation for having this on top of OpenConfig which already has a gNMI RPC layer?

This is the OpenConfig interfaces model that is mapped into a protobuf file.  Its essentially the same thing just made into a protobuf.  The gNMI can use the protobuf or any other orchestration client can use the protobuf format.

sandersms:
> I think we should add the Makefile logic and golang generated files before merging. Otherwise, looks good as a first start!

Added the Makefile and golang support to the protobuf.

mestery (changes requested):
Can you also run the `make` command check-in the golang generated files, as well as the autogen.md? This is how we do it in storage and security. Thanks!

mestery (approved):
We will want to checkin `autogen.md` as well. But we can do that in another commit, unless @sanders-mark adds it before merging this one.

PR #126, opened 2022-09-21
session: Add proxy ActionType

mestery:
This adds a new _PROXY ActionType to allow for offloading proxy sessions. Should work for both standard TCP and UDP proxies, as well as QUIC proxies.


mestery:
While I'm trying to solve offloading proxies here, the problem is that to get to the `SNI` value in a (D)TLS session or QUIC session requires us to handle the TCP handshake, thus the need to deal with sequence numbers on the socket. This assumes the socket is local to the application using these APIs. I was hoping to implement something like the Linux kernel eBPF Sockmap interface. That API assumes the application is running on the same kernel as the sockets. I think we can do the same thing here by requiring this API to be used by something running locally on the CPU complex of the DPU or IPU device. It's also possible I'm scratching an itch that no-one else has here, though I believe proxy offload is a valid use case.

glimchb (approved):
great!

sandersms (approved):
Looks like a good start on the proxy functions.  Lets move it forward.

PR #131, opened 2022-09-22
Storage: Update CSI related images

mkalderon:
Change terminology. Change Master nodes with Management nodes, to avoid harmful language.

Also added missing arrow in option #2


glimchb (approved):
I would prefer to switch between (1) and (2) gRPC commands but this PR is ok

PR #133, opened 2022-09-23
[WIP] nvme initiator protos

jainvipin:

Wanted to throw out the initiator protos to start the discussion. Tested the protoc compilation (and generation of pb.go files) from protoc (after emoving --lint) in the Makefile, all looks good there. but following pending
1. proto linting; move from CamelCase to under_score syntax; minor but required change
2. most importantly, need to merge the NVMe protos in the frontend, middleend, and backend hierarchy as it exists today

let's discuss what fits best for our needs and I will submit subsequent changes on #3 above
cc @glimchb @sandersms

### DIFF
--- a/storage/v1/CommonTypes.proto
+++ b/storage/v1/CommonTypes.proto
@@ -1,0 +1,174 @@
+//------------------------------------------------------------------------------
+// {C} Copyright 2022 Pensando Systems Inc. All rights reserved
+//
+// Common types for NVMe storage
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1.types;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+
+// API Status codes
+enum ApiStatusCode {
+  API_STATUS_OK                  = 0;
+  API_STATUS_OOM                 = 1;
+  API_STATUS_INVALID_ARG         = 2;
+  API_STATUS_INVALID_OP          = 3;
+  API_STATUS_ENTRY_NOT_FOUND     = 4;
+  API_STATUS_ENTRY_EXISTS        = 5;
+  API_STATUS_NO_RESOURCE         = 6;
+  API_STATUS_TABLE_FULL          = 7;
+  API_STATUS_OOB                 = 8;
+  API_STATUS_HW_PROGRAM_ERR      = 9;
+  API_STATUS_RETRY               = 10;
+  API_STATUS_NOOP                = 11;
+  API_STATUS_DUPLICATE_FREE      = 12;
+  API_STATUS_COLLISION           = 13;
+  API_STATUS_HW_READ_ERR         = 15;
+  API_STATUS_COMM_FAIL           = 19;
+  API_STATUS_IN_PROGRESS         = 22;
+  API_STATUS_TIMEOUT             = 39;
+  API_STATUS_ENTRY_READ_PARTIAL  = 40;
+  API_STATUS_ENTRY_BUSY          = 41;
+  API_STATUS_SVC_NOT_ACTIVE      = 43;
+  API_STATUS_OP_NOT_SUPPORTED    = 44;
+  API_STATUS_AGENT_NOT_ACTIVATED = 45;
+  API_STATUS_ERR                 = 255;
+}
+
+message ApiStatus {
+  ApiStatusCode Sc        = 1; // status code
+}
+
+// Global constants
+enum GlobalConstants {
+  NSV_NVME_CONST_NONE = 0; // dummy const
+  NSV_NVME_NQN_LEN    = 256;  // maximum length for NQN field
+}
+
+// Object Keys are 128bit (16byte) opaque values. The object key uniquely
+// identifies a given configuration object. The key used must be unique within
+// the agents context for the give object type and must be non-zero. No other
+// restrictions apply on the usage of the key. API user may utilize any number
+// of bits (upto 128) as required to uniquely identify the object. The same
+// object key must be provided in subsequent operations on object
+message ObjectKey {
+  bytes Key = 1;
+}
+
+// The controller PCI-ID is used to address a given virtual controller. Virtual
+// controllers are organized into devices with Physical functions and SRIOV
+// virtual function within the physical functions. Currently, the DSC may
+// expose multiple devices with one physical function each and one or more
+// virtual functions under the physical function.
+message NvmeControllerPciId {
+  // Bus number, provided for future usage if needed. Currently set to ’0’
+  uint32 Bus              = 1;
+  // Device number, based on the NVMe device layout
+  uint32 Device           = 2;
+  // Physical function, always set to 0 in current model
+  uint32 Function         = 3;
+  // SRIOV Virtual function within the Device and Physical function.
+  // Set to 0 for Physical Function. Virtual Function numbering starts from 1
+  uint32 VirtualFunction  = 4;
+}
+
+// AES encryption type to be used
+enum CryptType {
+  CRYPT_TYPE_NONE   = 0;  // must not be used
+  CRYPT_TYPE_AES128 = 1;  // Select AES 128bit encryption
+  CRYPT_TYPE_AES192 = 2;  // Select AES 192bit encryption
+  CRYPT_TYPE_AES256 = 3;  // Select AES 256bit encryption
+}
+
+// Path Management Selection
+enum NvmePathMgmtType {
+  PATH_MGMT_NONE            = 0; // none type, must not be used
+  PATH_MGMT_USER_MULTIPATH  = 1; // User managed multipath
+  PATH_MGMT_ANA_MULTIPATH   = 2; // NVMe Asymetric Namespace Access based multipath
+}
+
+// NVMe Fabric transport type
+enum NvmeTransportType {
+  TRANSPORT_TYPE_NONE = 0; // not specified, must not be used
+  TRANSPORT_TYPE_RDMA = 1; // RDMA transport
+  TRANSPORT_TYPE_FC   = 2; // Fibre Channel
+  TRANSPORT_TYPE_TCP  = 3; // TCP transport
+}
+
+// NVMe Address family in the NVMe/TCP port information.
+enum AddressFamily {
+  ADDRESS_FAMILY_NONE = 0;
+  ADDRESS_FAMILY_IPV4 = 1; // IPv4 (AF_INET)
+  ADDRESS_FAMILY_IPV6 = 2; // IPv6 (AF_INET6)
+  ADDRESS_FAMILY_IB   = 3; // InfiniBand (AF_IB)
+}
+
+// NVMe Fabric channel security considerations.
+enum NvmeFabricSecureChannelType {
+  SECURE_CHANNEL_NONE         = 0; // Not specified, must not be used
+  SECURE_CHANNEL_REQUIRED     = 1; // Secure channel required
+  SECURE_CHANNEL_NOT_REQUIRED = 2; // Secure channel not required
+}
+
+// TCP socket security
+enum NvmeTcpSecureSockType {
+  TCP_SECURITY_NONE = 0; // No Security
+  TCP_SECURITY_TLS  = 1; // TLS (Secure Sockets)
+}
+
+// TCP Additional Address information
+message NvmeTcpAdditionalInfo {
+  NvmeTcpSecureSockType SecureSockType = 1;
+}
+
+// RDMA Addtional Addr Info 
+message NvmeRdmaAdditionalInfo {
+  uint32 unused = 1;
+}
+
+// Transport Specific additional Address information
+message NvmeTransportAdditionalInfo {
+  oneof info {
+    NvmeTcpAdditionalInfo  Tcp  = 1;
+    NvmeRdmaAdditionalInfo Rdma = 2;
+  }
+}
+
+// Used to set/update the current operational mode of a path
+enum NvmePathMode {
+  // none, must not be used
+  PATH_MODE_NONE     = 0;
+  // Path is enabled, used to creating backed queues
+  PATH_MODE_ENABLED  = 1;
+  // Path is disabled, will not be used for creating backend queue
+  PATH_MODE_DISABLED = 2;
+}
+
+// Controller Type IDs for connection to fabric target
+enum NvmeControllerID {
+    CONTROLLER_ID_INVALID    = 0;      // not specified, must not be used
+    CONTROLLER_ID_STATIC_ANY = 0xFFFE; // Target controller is static type
+    CONTROLLER_ID_DYNAMIC    = 0xFFFF; // Target controller is dynamic type
+}
+
+// Nvme Namespace  Administrative States
+enum NvmeNamespaceState {
+  NAMESPACE_STATE_DISABLED    = 0;  // namespace disabled state
+  NAMESPACE_STATE_ENABLED     = 1;  // namespace enabled state
+  NAMESPACE_STATE_DELETING    = 2;  // namespace deleting state
+}
+
+// Namespace Operational States
+enum NvmeNamespaceOperState {
+  NAMESPACE_OPER_STATE_ONLINE  = 0;  // namespace is online
+  NAMESPACE_OPER_STATE_OFFLINE = 1;  // namespace is offline
+}
+
+// Path Connection States
+enum NvmePathConnectionState {
+  PATH_STATE_CONNECTING    = 0; // path connecting state
+  PATH_STATE_CONNECTED     = 1; // path connected state
+  PATH_STATE_DISCONNECTING = 2; // path disconnecting state
+  PATH_STATE_DISCONNECTED  = 3; // path disconnected state
+}
--- a/storage/v1/CommonTypes.proto
+++ b/storage/v1/CommonTypes.proto
@@ -58,7 +58,7 @@ message ObjectKey {
 
 // The controller PCI-ID is used to address a given virtual controller. Virtual
 // controllers are organized into devices with Physical functions and SRIOV
-// virtual function within the physical functions. Currently, the DSC may
+// virtual function within the physical functions. An xPU may
 // expose multiple devices with one physical function each and one or more
 // virtual functions under the physical function.
 message NvmeControllerPciId {
--- a/storage/v1/CommonTypes.proto
+++ b/storage/v1/CommonTypes.proto
@@ -37,7 +37,7 @@ enum ApiStatusCode {
 }
 
 message ApiStatus {
-  ApiStatusCode Sc        = 1; // status code
+  ApiStatusCode sc        = 1; // status code
 }
 
 // Global constants
@@ -53,7 +53,7 @@ enum GlobalConstants {
 // of bits (upto 128) as required to uniquely identify the object. The same
 // object key must be provided in subsequent operations on object
 message ObjectKey {
-  bytes Key = 1;
+  bytes key = 1;
 }
 
 // The controller PCI-ID is used to address a given virtual controller. Virtual
@@ -63,14 +63,14 @@ message ObjectKey {
 // virtual functions under the physical function.
 message NvmeControllerPciId {
   // Bus number, provided for future usage if needed. Currently set to ’0’
-  uint32 Bus              = 1;
+  uint32 bus              = 1;
   // Device number, based on the NVMe device layout
-  uint32 Device           = 2;
+  uint32 device           = 2;
   // Physical function, always set to 0 in current model
-  uint32 Function         = 3;
+  uint32 function         = 3;
   // SRIOV Virtual function within the Device and Physical function.
   // Set to 0 for Physical Function. Virtual Function numbering starts from 1
-  uint32 VirtualFunction  = 4;
+  uint32 virtual_function  = 4;
 }
 
 // AES encryption type to be used
@@ -119,7 +119,7 @@ enum NvmeTcpSecureSockType {
 
 // TCP Additional Address information
 message NvmeTcpAdditionalInfo {
-  NvmeTcpSecureSockType SecureSockType = 1;
+  NvmeTcpSecureSockType secure_sock_type = 1;
 }
 
 // RDMA Addtional Addr Info 
@@ -130,8 +130,8 @@ message NvmeRdmaAdditionalInfo {
 // Transport Specific additional Address information
 message NvmeTransportAdditionalInfo {
   oneof info {
-    NvmeTcpAdditionalInfo  Tcp  = 1;
-    NvmeRdmaAdditionalInfo Rdma = 2;
+    NvmeTcpAdditionalInfo  tcp  = 1;
+    NvmeRdmaAdditionalInfo rdma = 2;
   }
 }
 
--- a/storage/v1/NvmeIniDev.proto
+++ b/storage/v1/NvmeIniDev.proto
@@ -19,12 +19,12 @@ service NvmeIniDevSvc {
 // NvmeIni device specification
 message NvmeIniDevSpec {
   // (mandatory) initiator device object key.
-  types.ObjectKey NvmeIniDevKey    = 1;
+  types.ObjectKey nvme_ini_dev_key    = 1;
   // (mandatory) unique NQN identifer for the host, used in fabric connect to
   // the NVMe-oF target
-  bytes        HostNqn             = 2;
+  bytes        host_nqn             = 2;
   // (mandatory) unique host ID, used in fabric connect to the NVMe-oF target.
-  bytes        HostId              = 3;
+  bytes        host_id              = 3;
 }
 
 // initiator device status
@@ -37,38 +37,38 @@ message NvmeIniDevStats {
 
 // initiator device object
 message NvmeIniDev {
-  NvmeIniDevSpec   Spec   = 1; // specification
-  NvmeIniDevStatus Status = 2; // status
-  NvmeIniDevStats  Stats  = 3; // stats
+  NvmeIniDevSpec   spec   = 1; // specification
+  NvmeIniDevStatus status = 2; // status
+  NvmeIniDevStats  stats  = 3; // stats
 }
 
 // create request
 message NvmeIniDevCreateRequest {
-  NvmeIniDevSpec Spec = 1; // initiator object specification
+  NvmeIniDevSpec spec = 1; // initiator object specification
 }
 
 // create response
 message NvmeIniDevCreateResponse {
-  types.ApiStatus    ApiStatus = 1; // status code
+  types.ApiStatus    api_status = 1; // status code
 }
 
 // get request
 message NvmeIniDevGetRequest {
-  types.ObjectKey NvmeIniDevKey = 1; // object to get
+  types.ObjectKey nvme_ini_dev_key = 1; // object to get
 }
 
 // get response
 message NvmeIniDevGetResponse {
-  types.ApiStatus ApiStatus = 1; // status code
-  NvmeIniDev      Response  = 2; // iniitiator device object
+  types.ApiStatus api_status = 1; // status code
+  NvmeIniDev      response  = 2; // iniitiator device object
 }
 
 // delete Request
 message NvmeIniDevDeleteRequest {
-  types.ObjectKey NvmeIniDevKey = 1; // initiator device object key
+  types.ObjectKey nvme_ini_dev_key = 1; // initiator device object key
 }
 
 // delete response
 message NvmeIniDevDeleteResponse {
-  types.ApiStatus    ApiStatus = 1; // status code
+  types.ApiStatus    api_status = 1; // status code
 }
--- a/storage/v1/NvmeIniDev.proto
+++ b/storage/v1/NvmeIniDev.proto
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 // {C} Copyright 2022 Pensando Systems Inc. All rights reserved
 //
-// Nvme iniaiator device proto specification
+// Nvme Initiator device proto specification
 //------------------------------------------------------------------------------
 
 syntax = "proto3";
--- a/storage/v1/NvmeIniDev.proto
+++ b/storage/v1/NvmeIniDev.proto
@@ -1,0 +1,74 @@
+//------------------------------------------------------------------------------
+// {C} Copyright 2022 Pensando Systems Inc. All rights reserved
+//
+// Nvme iniaiator device proto specification
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+// initiator device service
+service NvmeIniDevSvc {
+  rpc NvmeIniDevCreate (NvmeIniDevCreateRequest) returns (NvmeIniDevCreateResponse) {}
+  rpc NvmeIniDevGet (NvmeIniDevGetRequest) returns (NvmeIniDevGetResponse) {}
+  rpc NvmeIniDevDelete (NvmeIniDevDeleteRequest) returns (NvmeIniDevDeleteResponse) {}
+}
+
+// NvmeIni device specification
+message NvmeIniDevSpec {
+  // (mandatory) initiator device object key.
+  types.ObjectKey NvmeIniDevKey    = 1;
+  // (mandatory) unique NQN identifer for the host, used in fabric connect to
+  // the NVMe-oF target
+  bytes        HostNqn             = 2;
+  // (mandatory) unique host ID, used in fabric connect to the NVMe-oF target.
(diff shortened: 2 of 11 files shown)


glimchb (changes requested):
@jainvipin this is excellent
as agreed, we will work together to break this commit into several chunks and merge into existing APIs !
appreciate the contribution

jainvipin:
> @jainvipin this is excellent as agreed, we will work together to break this commit into several chunks and merge into existing APIs ! appreciate the contribution

sounds good! I fixed lint errors in the meanwhile and verified all code generation is looking good. let's discuss on the merge!

glimchb:
@orendu and @GottliebNoam , please review this once the official holidays are over and you back in the office

glimchb:
@jainvipin 

I think we can start with merging proposed `storage/proto/NvmeControllerPci.proto` 
into `NVMeControllerService` from https://github.com/opiproject/opi-api/blob/main/storage/proto/frontend-nvme-pcie.proto 

I see 
```
service NvmeControllerSvc {
  // Instantiate a new controller on the given PCI-ID
  rpc NvmeControllerCreate (NvmeControllerCreateRequest) returns (NvmeControllerCreateResponse) {}
  // Enable a namespace attached to the controller
  rpc NvmeControllerEnableNs (NvmeControllerEnableNsRequest) returns (NvmeControllerEnableNsResponse) {}
  // Disable a namespace attached to the controller
  rpc NvmeControllerDisableNs (NvmeControllerDisableNsRequest) returns (NvmeControllerDisableNsResponse) {}
  // Get controller information
  rpc NvmeControllerGet (NvmeControllerGetRequest) returns (NvmeControllerGetResponse) {}
  // Delete a controller
  rpc NvmeControllerDelete (NvmeControllerDeleteRequest) returns (NvmeControllerDeleteResponse) {}
}
```
vs
```
service NVMeControllerService {
    rpc NVMeControllerCreate (NVMeControllerCreateRequest) returns (NVMeControllerCreateResponse) {}
    rpc NVMeControllerDelete (NVMeControllerDeleteRequest) returns (NVMeControllerDeleteResponse) {}
    rpc NVMeControllerUpdate (NVMeControllerUpdateRequest) returns (NVMeControllerUpdateResponse) {}
    rpc NVMeControllerList   (NVMeControllerListRequest)   returns (NVMeControllerListResponse)   {}
    rpc NVMeControllerGet    (NVMeControllerGetRequest)    returns (NVMeControllerGetResponse)    {}
    rpc NVMeControllerStats  (NVMeControllerStatsRequest)  returns (NVMeControllerStatsResponse)  {}
}
```

sounds reasonable ? WDYT ?

storage/proto/CommonTypes.proto
```diff
@@ -0,0 +1,174 @@
...
+// of bits (upto 128) as required to uniquely identify the object. The same
+// object key must be provided in subsequent operations on object
+message ObjectKey {
+  bytes key = 1;
+}
+
+// The controller PCI-ID is used to address a given virtual controller. Virtual
+// controllers are organized into devices with Physical functions and SRIOV
+// virtual function within the physical functions. Currently, the DSC may
+// expose multiple devices with one physical function each and one or more
```
glimchb, line 62: there are few places mentioning `DSC` specifically

jainvipin: Ack.

storage/proto/NvmeControllerPci.proto
```diff
@@ -0,0 +1,149 @@
...
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+service NvmeControllerSvc {
+  // Instantiate a new controller on the given PCI-ID
+  rpc NvmeControllerCreate (NvmeControllerCreateRequest) returns (NvmeControllerCreateResponse) {}
+  // Enable a namespace attached to the controller
+  rpc NvmeControllerEnableNs (NvmeControllerEnableNsRequest) returns (NvmeControllerEnableNsResponse) {}
```
glimchb, line 16: I'm thinking about `NvmeControllerEnableNs` and `NvmeControllerDisableNs` now...
do they belong to the `NvmeControllerSvc` or to the `NvmeNamespacePciSvc` ?

glimchb:
I see `storage/proto/NvmeControllerPci.proto` and `storage/proto/NvmeNamespacePci.proto` to the HOST, but where is `storage/proto/NvmeSubsystemPci.proto` ?

llabordehpe:
@jainvipin : two questions about your API:
- Why are the object handles assigned by the client? How is the client suppoed to know what is OK for an opaque handle? Considering that the OPI API is supposed to work will all vendors/IPUs, I think the opaque handle has to be assigned by the provider.
- Why is there a field "types.ApiStatus  api_status" in your protobuf response messages ? Don't you get your operation status from the gRPC status codes ?

storage/proto/NvmeTargetPort.proto
```diff
@@ -0,0 +1,82 @@
...
+//
+// Protobuf specification for NVMe target port
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+// target port service
```
glimchb, line 12: this looks more like target APIs and not Initiator APIs, true ?

storage/proto/NvmeTargetSubsys.proto
```diff
@@ -0,0 +1,80 @@
...
+// Protobuf specification for NVMe target subsys APIs
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+// target subsys service
+service NvmeTargetSubsystemSvc {
```
glimchb, line 13: this looks more like target APIs and not Initiator APIs, true ?

jainvipin: this is the target configuration on the Initiator, so you can say these are target (or back-end) APIs based on how you have divided the APIs to be front-end, middle-end, and the back-end in your classification.

storage/proto/NvmeTargetPort.proto
jainvipin: I am calling all these initiator APIs because of where these APIs are implemented, but the target port is a configuration on the Initiator.

jainvipin:
> @jainvipin : two questions about your API:
> 
> * Why are the object handles assigned by the client? How is the client suppoed to know what is OK for an opaque handle? Considering that the OPI API is supposed to work will all vendors/IPUs, I think the opaque handle has to be assigned by the provider.
> * Why is there a field "types.ApiStatus  api_status" in your protobuf response messages ? Don't you get your operation status from the gRPC status codes ?

@llabordehpe - appreciate your observation on this. The way I see, there are two ways of uniquely identifying an object. 1) based on the key fields (e.g. target-ip, port, etc.), in this case a handle is assigned by the provider. 2) based on a handle e.g. name or uuid (which is also a special type of string), provider can still return its own handle in this case as well, but can use the name/uuid provided by the client to organize its lookups. The benefit of second approach is that it can differentiate the instances of two objects with same key fields as two instances of an object, since the uuids can change but key fields would be same. e.g. if we want to introduce (target-ip,port) as unique set of fields and would like to introduce tenant-id, or sub-systemid later we can do that without altering the semantics of the APIs.

Also, given that `Spec` fields can change and evolve, in these APIs we preferred going with approach #2, so we have flexibility to evolve the APIs without changing the uniqueness semantics. Use of UUID is not too uncommon to identify a object instance. K8s APIs for example use a name to uniquely identify an object, which can be a uuid string if instance uniqueness is required.

In either cases, backend implementation is easier if it uses an id/handle based lookup than searching on the key fields. So the current APIs have a unique `[] byte` opaque key field for various objects.

storage/proto/NvmeIniDev.proto
```diff
@@ -0,0 +1,74 @@
...
+  rpc NvmeIniDevCreate (NvmeIniDevCreateRequest) returns (NvmeIniDevCreateResponse) {}
+  rpc NvmeIniDevGet (NvmeIniDevGetRequest) returns (NvmeIniDevGetResponse) {}
+  rpc NvmeIniDevDelete (NvmeIniDevDeleteRequest) returns (NvmeIniDevDeleteResponse) {}
+}
+
+// NvmeIni device specification
+message NvmeIniDevSpec {
+  // (mandatory) initiator device object key.
+  types.ObjectKey nvme_ini_dev_key    = 1;
+  // (mandatory) unique NQN identifer for the host, used in fabric connect to
```
jkogutINTC, line 23: Is it really the same NQN identified exposed to host which is used in connection to the NVMe-oF target?

storage/proto/NvmeIniDev.proto
```diff
@@ -0,0 +1,74 @@
...
+//
+// Nvme iniaiator device proto specification
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+// initiator device service
```
jkogutINTC, line 12: it will be nice to have short overview in comment what it is "Nvme iniaiator device"? It is a description how the device will be exposed to host?

The same comment is for other introduced ne definitions in the review.

storage/proto/NvmeIniDev.proto
```diff
@@ -0,0 +1,74 @@
+//------------------------------------------------------------------------------
+// {C} Copyright 2022 Pensando Systems Inc. All rights reserved
+//
+// Nvme iniaiator device proto specification
```
jkogutINTC, line 4: spelling iniaiator -> initiator

storage/proto/NvmeTargetSubsys.proto
glimchb: I thought that `storage/proto/NvmePath.proto` is  back-end (network facing APIs)...

storage/proto/CommonTypes.proto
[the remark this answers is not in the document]
glimchb: +1. should be called `DPU/IPU`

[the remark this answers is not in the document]
jainvipin: sure, will call it xPU

storage/proto/NvmeIniDev.proto
jainvipin: fixed

storage/proto/CommonTypes.proto
```diff
@@ -0,0 +1,174 @@
...
+// virtual functions under the physical function.
```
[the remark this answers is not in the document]
orendu: +1
B:D:F already defines any PCI function, physical or virtual.
Not clear what is the purpose of this virtual_function field.

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+//
+// Protobuf specification for NVMe namespace service
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+service NvmeNamespacePciSvc {
```
orendu, line 12: What is the meaning of "Pci" in all the names related to namespace management?
A namespace is an NVMe object (of a subsystem) unrelated to PCI or any other transport...

storage/v1/CommonTypes.proto
```diff
@@ -0,0 +1,174 @@
...
+// controllers are organized into devices with Physical functions and SRIOV
+// virtual function within the physical functions. An xPU may
+// expose multiple devices with one physical function each and one or more
+// virtual functions under the physical function.
+message NvmeControllerPciId {
+  // Bus number, provided for future usage if needed. Currently set to ’0’
+  uint32 bus              = 1;
+  // Device number, based on the NVMe device layout
+  uint32 device           = 2;
+  // Physical function, always set to 0 in current model
```
orendu, line 69: Comment about always set to 0 should be removed, it is vendor choice.

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+  // If multiple namespaces share the same path, then the value provided in the
+  // first namespace in the share group will be used.
+  uint32             retry_interval       = 3;
+  // (optional) flag to control when backend session connection is established.
+  // 1:  auto connect backend session on namespace enable. backend IO
+  // connections are created irrespective of state of the frontend controller
+  // object i.e enabled/disabled by the host driver.
+  // 0: connect backend session only after controller is enabled by host
+  // driver.
+  bool               auto_connect         = 4;
```
orendu, line 37: In the layering we discussed for OPI, this should be part of the backend APIs, not the frontend.
I.e., in SPDK, when NVMe-oF connection is created to a remote subsystem, bdevs are created as a result. This is the right place to indicate auto_connect, and retry_interval.

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+  rpc NvmeNamespacePciUpdate (NvmeNamespacePciUpdateRequest) returns (NvmeNamespacePciUpdateResponse) {}
+}
+
+message NvmeNamespacePciParams {
+  // (mandatory) NVMe NSID of the namespace in the backend target subsystem
+  // hosting this namespace.
+  uint32             tgt_nsid             = 1;
+  // (mandatory) Size/Capacity of the namespace in blocks, size in bytes will
+  // be BlockSize x NumBlocks.
+  uint64             num_blocks           = 2;
```
orendu, line 25: This should probably be taken from the backend object assigned to this namespace.

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+import "CommonTypes.proto";
+
+service NvmeNamespacePciSvc {
+  rpc NvmeNamespacePciCreate (NvmeNamespacePciCreateRequest) returns (NvmeNamespacePciCreateResponse) {}
+  rpc NvmeNamespacePciGet (NvmeNamespacePciGetRequest) returns (NvmeNamespacePciGetResponse) {}
+  rpc NvmeNamespacePciDelete (NvmeNamespacePciDeleteRequest) returns (NvmeNamespacePciDeleteResponse) {}
+  rpc NvmeNamespacePciUpdate (NvmeNamespacePciUpdateRequest) returns (NvmeNamespacePciUpdateResponse) {}
+}
+
+message NvmeNamespacePciParams {
```
orendu, line 19: Missing a reference to the backend object representing the backend volume assigned to this namespace.

storage/v1/NvmeNsCrypto.proto
```diff
@@ -0,0 +1,79 @@
...
+//
+// Crypto support for NVMe Namespace
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+service NvmeNsCryptoSvc {
```
orendu, line 12: According to OPI storage API layering, crypto should be part of the middle-end APIs.
i.e., take any backend volume object (created by backend API), define crypto on top of it (middle-end API), then use the result in a frontend NVMe namespace or virtio-blk volume (frontend APIs).

orendu:
Could we see an example flow of expected calls to this API that will eventually set up a remote NVMe-oF namespace as a local PCI NVMe controller and namespace?

It will help understanding the full picture.

I suspect that some of the APIs here belong to backend APIs. NvmeTargetSubsys (if I got it correctly), NVMeTargetPort, NvmePath - are they all to define a connection (with multipath) to remote storage target? If so, they belong to the backend APIs.

Crypto should be middle-end.

Missing the concept of a subsystem in the frontend (with multiple controllers and namespaces in it).

storage/v1/CommonTypes.proto
```diff
@@ -0,0 +1,174 @@
...
+// object key must be provided in subsequent operations on object
+message ObjectKey {
+  bytes key = 1;
+}
+
+// The controller PCI-ID is used to address a given virtual controller. Virtual
+// controllers are organized into devices with Physical functions and SRIOV
+// virtual function within the physical functions. An xPU may
+// expose multiple devices with one physical function each and one or more
+// virtual functions under the physical function.
```
orendu, line 63: An xPU may also expose one device with multiple physical functions...

storage/v1/CommonTypes.proto
```diff
@@ -0,0 +1,174 @@
...
+message NvmeControllerPciId {
+  // Bus number, provided for future usage if needed. Currently set to ’0’
+  uint32 bus              = 1;
+  // Device number, based on the NVMe device layout
+  uint32 device           = 2;
+  // Physical function, always set to 0 in current model
+  uint32 function         = 3;
+  // SRIOV Virtual function within the Device and Physical function.
+  // Set to 0 for Physical Function. Virtual Function numbering starts from 1
+  uint32 virtual_function  = 4;
```
orendu, line 73: As pointed in another comment, the triplet {bus:device:function} is enough to describe both virtual and physical functions. What is the need for this additional field?

intelfisz:
> Could we see an example flow of expected calls to this API that will eventually set up a remote NVMe-oF namespace as a local PCI NVMe controller and namespace?
> 
> It will help understanding the full picture.
> 
> I suspect that some of the APIs here belong to backend APIs. NvmeTargetSubsys (if I got it correctly), NVMeTargetPort, NvmePath - are they all to define a connection (with multipath) to remote storage target? If so, they belong to the backend APIs.
> 
> Crypto should be middle-end.
> 
> Missing the concept of a subsystem in the frontend (with multiple controllers and namespaces in it).

Thanks @orendu for pointing it out. This is exactly my point - for such a comprehensive commit we need at least some high-level architecture and/or usage examples - can you provide that @jainvipin? @glimchb can you elaborate on the expected main clients in the [Common APIs abstraction layer](https://github.com/opiproject/opi-api/tree/main/storage#common-apis-abstraction-layer)? I assume K8s nodes.

storage/v1/NvmeNamespacePci.proto
glimchb: I guess this is NVMe/PCI emulation presented to the host, this is why Pci is here...

glimchb:
> Missing the concept of a subsystem in the frontend (with multiple controllers and namespaces in it).
@orendu I also pointed this to @jainvipin 
he answered me privately, maybe he can paste his answer here as well...

glimchb:
> @glimchb can you elaborate on the expected main clients in the [Common APIs abstraction layer](https://github.com/opiproject/opi-api/tree/main/storage#common-apis-abstraction-layer)? I assume K8s nodes.

@intelfisz  one of the clients can be k8s via csi. we track this work here https://github.com/opiproject/goopicsi and here https://github.com/opiproject/opi-api/blob/main/storage/opi-csi.md

But we also have other orchestrators, non k8s based that can send the commands to configure many of DPU/IPUs in the datacenter.

storage/v1/NvmeNamespacePci.proto
jainvipin: Host facing is named suffixed with `pci`

storage/proto/NvmeIniDev.proto
jainvipin: all host exposed fields are in `NvmeControllerPci` object, therefore this nqn is used to identiy the device to connect to the NVMe-oF target (which would have configured to accept connections from this nqn)

storage/v1/CommonTypes.proto
jainvipin: ack, will fix.

storage/v1/NvmeNamespacePci.proto
jainvipin: configuring this here can allow an abstraction of showing different size. we can make this optional - would that work?

storage/v1/NvmeNamespacePci.proto
jainvipin: let's discuss; auto-connect is meant to allow the namespace to be primed before NvmeController is enabled on the host - this is specific to namespace not the backend config (path/subsystem objects in this PR), which is shared by multiple namespaces and thus not controllable per namespace. This PR doesn't have a bdev backend object, and so.

jainvipin:
> Could we see an example flow of expected calls to this API that will eventually set up a remote NVMe-oF namespace as a local PCI NVMe controller and namespace?

yes, will document example flow. 

> It will help understanding the full picture.
> 
> I suspect that some of the APIs here belong to backend APIs. NvmeTargetSubsys (if I got it correctly), NVMeTargetPort, NvmePath - are they all to define a connection (with multipath) to remote storage target? If so, they belong to the backend APIs.

yes there are backend, frontend and middleend apis as part of this PR - but for NVMe.

> Crypto should be middle-end.
> 
> Missing the concept of a subsystem in the frontend (with multiple controllers and namespaces in it).

@glimchb also made this observation. we have the configuration in the controller object, and subsystem nqn is included there too. However, we can keep the subsystem hierarchy explicitly.

jainvipin:
> @orendu @intelfisz @jainvipin we can definitely take this PR and split in smaller pieces and review and merge them... I already did one small chunk here #134

sounds like a good idea. we'll submit smaller consumable PRs out of comments in this PR and go with that.

llabordehpe:
> 

@jainvipin: The general gRPC error handling (https://grpc.io/docs/guides/error/), and particularly the richer error model (https://cloud.google.com/apis/design/errors#error_model) should have everything you need to convey errors to the client, without defining new status/error fields in your protobuf response messages.

yadong2007:
> 

Hi, there is an interesting proposal.  It enables discussions.
After I took a quick look at the proposal, I feel it's hard for me to understand the exact level of abstraction this proposal tries to achieve.  A few comments and questions:
- First, it's not clear what exactly the "Client" that you are trying to enable.  If it's K8s CSI node driver, then exposing a lot of details in NVMe subsystem and NVMe controllers do not help simplify anything.  It will be good to clear state the "client", and then drive top level API definitions from there.
- Second, it's also not clear what exactly the Vendor SDK or IPU/DPU SDK you are trying to enable.  If it's SPDK implementation, then a lot of the APIs in this proposal seems duplicate or similar to SPDK JSON RPCs.
- To make the bridge/shim API defined properly, I think we need start with a "Client" and a "IPU/DPK SDK".  To make sure it works at least for one of such clients.
- I also highly recommend splitting the API definition for different usages,  "Initiator" vs "Target".  The APIs for Initiators can be highly simplified.  The Target API is often tied to vendor specific storage server design, it will take time to get to a standard based API definition.

glimchb:
@yadong2007 did you look at the existing already merged APIs https://github.com/opiproject/opi-api/tree/main/storage/v1 ?
also check out OPI-SPDK bridge we built https://github.com/opiproject/opi-poc/tree/main/storage/server
also check CSI library we building https://github.com/opiproject/goopicsi based on https://github.com/opiproject/opi-api/blob/main/storage/opi-csi.md

yadong2007:
> @yadong2007 did you look at the existing already merged APIs https://github.com/opiproject/opi-api/tree/main/storage/v1 ? also check out OPI-SPDK bridge we built https://github.com/opiproject/opi-poc/tree/main/storage/server also check CSI library we building https://github.com/opiproject/goopicsi based on https://github.com/opiproject/opi-api/blob/main/storage/opi-csi.md

@glimchb Hi Boris, thanks for the links.  I remember these patches.  It's a very good starting point. 
I like the architecture diagrams in this page: https://github.com/opiproject/opi-api/blob/main/storage/opi-csi.md  I think it's more effective to illustrate the key API call sequences using CSI node driver as an example client.   Basically adding a few sequence diagrams will make it clear what the exact APIs exposed to the CSI node driver and to the CSI controller driver in the management node.

glimchb:
@yadong2007 I agree. can you help out and create some sequence diagrams so we can review them ?

yadong2007:
> @yadong2007 I agree. can you help out and create some sequence diagrams so we can review them ?

Hi Boris, yes, Intel team is working on the sequence diagrams.  Will share whenever it's ready in next couple of weeks.

glimchb:
> > @yadong2007 I agree. can you help out and create some sequence diagrams so we can review them ?
> 
> Hi Boris, yes, Intel team is working on the sequence diagrams. Will share whenever it's ready in next couple of weeks.

I created one here, let me know what you think https://github.com/opiproject/opi-poc/pull/387/files

storage/proto/CommonTypes.proto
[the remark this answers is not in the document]
intelfisz: @pwalessi-dell this also needs to be aligned here: https://github.com/opiproject/opi-api/blob/main/storage/v1/common.proto

jainvipin:
> > 
> 
> @jainvipin: The general gRPC error handling (https://grpc.io/docs/guides/error/), and particularly the richer error model (https://cloud.google.com/apis/design/errors#error_model) should have everything you need to convey errors to the client, without defining new status/error fields in your protobuf response messages.

@llabordehpe - I understand grpc is very flexible on what error codes can return; however we were unable to capture the details e.g. Nvme specific fields/values to be returned alongwith details to referring objects and transport layer is not capturing it. Would it not allow us to keep better flexibility and debuggability better for NVMe layer if we have a verbose status reponse?

storage/proto/NvmeTargetSubsys.proto
jainvipin: yes, will move these to backend protos @glimchb in a separate PR for backend objects.

storage/proto/NvmeControllerPci.proto
```diff
@@ -0,0 +1,149 @@
...
+  // Must not be reused under the same subsystem
+  uint32 nvme_controller_id  = 1;
+  // (mandatory) Serial Number must not exceed 'NSV_CTRLR_SERIAL_NO_LEN' bytes
+  string serial_number      = 2;
+  // (mandatory) Model Number, must not exceed 'NSV_CTRLR_MODEL_NO_LEN' bytes
+  string model_number       = 3;
+  // (mandatory) Firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
+  string firmware_revision  = 4;
+  // (mandatory) FRU global identfier, 16bytes opaque identity for the unit
+  bytes  fru_guid           = 5;
```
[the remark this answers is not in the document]
jainvipin: @jkogutINTC - are you suggesting we change the name to FGUID or the size to be specific, the idea was that bytes would cover this to be 16bytes value.

storage/v1/NvmeNamespacePci.proto
jainvipin: @orendu - will put a diagram to explain the backend objects a bit, but it is indirectly referenced through path objects.

storage/v1/NvmeIniDev.proto
```diff
@@ -0,0 +1,74 @@
...
+// Nvme Initiator device proto specification
+//------------------------------------------------------------------------------
+
+syntax = "proto3";
+package opi.storage.v1;
+option go_package = "github.com/opiproject/opi-api/storage/proto";
+import "CommonTypes.proto";
+
+// initiator device service
+service NvmeIniDevSvc {
```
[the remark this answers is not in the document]
jainvipin: @orendu this is the initiator personality (nqn) to identify itself to the backend

storage/v1/CommonTypes.proto
jainvipin: hmm... haven't heard of this use case; @glimchb @llabordehpe - any comments on this use case?

storage/v1/CommonTypes.proto
jainvipin: @orendu - let me get back on this one and update this comment later

jainvipin:
@yadong2007 - some reponses inline..
 
> Hi, this is an interesting proposal. It enables discussions. After I took a quick look at the proposal, I feel it's hard for me to understand the exact level of abstraction this proposal tries to achieve. A few comments and questions:
> 
> * First, it's not clear what exactly the "Client" that you are trying to enable.  If it's K8s CSI node driver, then exposing a lot of details in NVMe subsystem and NVMe controllers do not help simplify anything.  It will be good to clearly state the "client", and then drive top level API definitions from there.

The client is really a configurator of the Nvme devices on the DPU; it doesn't even need to be residing on the host e.g. BM provisioning will keep this API to be called from over the cloud infrastructure network. Having said that,  this can be used by a CSI driver in k8s deployment to provision volumes and bind mount them into pods as they get scheduled on nodes.

> * Second, it's also not clear what exactly the Vendor SDK or IPU/DPU SDK you are trying to enable.  If it's SPDK implementation, then a lot of the APIs in this proposal seems duplicate or similar to SPDK JSON RPCs.

The backend of these APIs could be SPDK or a P4 implementation. The reason to propose these APIs is to have the flexibility to use different implementations for various platform and capabilities.

> * To make the bridge/shim API defined properly, I think we need start with a "Client" and a "IPU/DPU SDK".  To make sure it works at least for one of such clients.
Of course. The APIs are going to be  (or have already been) vetted against a working implementation.

> * I also highly recommend splitting the API definition for different usages,  "Initiator" vs "Target".  The APIs for Initiators can be highly simplified.  The Target API is often tied to vendor specific storage server design, it will take time to get to a standard based API definition.

Good point; I will change this; this is creating a good amount of confusion to many reviewers.

storage/v1/NvmeNamespacePci.proto
orendu: So you mean this is a flag indicating that a specific namespace must be present before the controller is exposed to the host?
For example, a controller is made of two namespaces, and one of them does not exist at the moment (remote is down, for instance).  If the flag is set - enabling the controller will be blocked, and if it is unset, enabling the controller with be allowed and the missing namespace won't be presented?
If this is a correct description, then maybe the flag should be named "mandatory" or alike...

storage/v1/NvmeNamespacePci.proto
orendu: yes, it could act as some kind of "override" to the properties of the underlying backend. block_size could also potentially be changed...
Also consider, its may be natural to explain what happens if the override reduces the total capacity of the backend volume, but what happens if it increases it?

storage/v1/NvmeIniDev.proto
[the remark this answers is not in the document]
glimchb: is this equivalent to https://spdk.io/doc/jsonrpc.html#rpc_nvmf_subsystem_add_host ?

storage/v1/CommonTypes.proto
mkalderon: @orendu , @jainvipin aren't you referring to the same ? multi-function devices ? 
@jainvipin what do you mean by xPU exposes mutiple devices - don't you mean multiple pfs ?

storage/v1/CommonTypes.proto
mkalderon: maybe has to do with mutli-function device discussed above, but xPU could expose several PFs where Function won't be 0

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+  uint32             pref_write_granularity= 6;
+}
+
+// namespace specification
+message NvmeNamespacePciSpec {
+  // (mandatory) key for the namespace object.
+  types.ObjectKey    nvme_namespace_pci_key = 1;
+  // (mandatory) key of the PCIe controller object that will host this
+  // namespace.
+  types.ObjectKey    controller_key       = 2;
```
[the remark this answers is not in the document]
glimchb: in dual ported NVMe drives you have 1 subsystem with 2 controllers... 
and NSID is attached to controllers either privately or shared between multiple controllers... 
no ?

see

storage/v1/NvmeNamespacePci.proto
```diff
@@ -0,0 +1,168 @@
...
+  // (mandatory) Encryption key for the data at rest encryption.
+  types.ObjectKey    crypto_key_key        = 3;
+  // (optional) Cluster object to which the corresponding backend namespace
+  // belongs. If not provided, then namespace will be assigned to default
+  // cluster.
+  types.ObjectKey    cluster_key          = 4;
+  // (mandatory) Initiator device for the namespace, corresponding initiator
```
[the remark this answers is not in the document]
glimchb: +1

storage/v1/NvmeNamespacePci.proto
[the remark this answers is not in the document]
benlwalker: Correct that's the only case where they're not 1:1 for a local PCIe SSD. If we feel it is necessary to be able to model that in an xPU we can discuss that. But I doubt we need to.

jainvipin:
closing this pr in favor of the work that has gone in-parts to the storage apis; thanks everyone for the review, comments and support

PR #135, opened 2022-09-28
Add aip linter action

PWAlessi:
Action does not fail build, it just outputs results


glimchb (commented):
this looks good!
if we merge it, will our CI starts failing because our APIs are not passing the linter ?

PR #136, opened 2022-09-28
unify linters to match other repos we have

glimchb:


mestery (approved):
This is fine, but make sure you also update the branch protection jobs once this merges to require the new jobs.

glimchb:
> This is fine, but make sure you also update the branch protection jobs once this merges to require the new jobs.

protection cares only about job name, not workflow name or file name, so old definition of `markdown-lint` still correct
added new `api-lint` to branch protection

PR #139, opened 2022-09-29
proto: add python and cpp

glimchb:
- remove auto-gen files
- proto: add python and cpp


network/Makefile
```diff
@@ -5,5 +5,7 @@
 all:
 	rm -rf ./proto/{autogen.md,*.pb.go}
 	docker run -v "${PWD}":/defs namely/protoc-all:1.47_2 --lint -d proto -l go -o ./proto/  --go-source-relative
```
mestery, line 7: Can we do this in one go with commas between generated versions? I'm fine either way.

mestery (approved):
Yup!

network/Makefile
glimchb: I wanted separate commands to maybe use separate folders for generated files later on... 
will dig into documentation a bit...

PR #144, opened 2022-10-03
Implement File and directory structure per AIP-191

PWAlessi:
Implement File and directory structure per AIP-191 https://google.aip.dev/191

Renamed files to use snake_case
Moved protos into v1 directories
Updated package declarations to use correct directory path with snake_case
Updated go_package directory locations
Updated makefiles

Ultimately, we should remove the existing built artifacts from the proto directories too. I'm leaving them in place until we update the POC to build the proto artifacts itself.


glimchb:
@llabordehpe @jainvipin FYI

glimchb (commented):
@PWAlessi 
the folder structure is better to accommodate versioning, thanks for that
for package name: why do you prefer `opi_api` instead of simply `opi` ?

PWAlessi:
> @PWAlessi the folder structure is better to accommodate versioning, thanks for that for package name: why do you prefer `opi_api` instead of simply `opi` ?

opi_api matches the directory structure

PWAlessi:
I've just committed breaking changes that move the generated artifacts below the proto definitions and into language specific directories.

glimchb:
@PWAlessi this is great, please update the linter as well so we pass checkers
see `--ignore` here https://github.com/opiproject/opi-api/blob/main/.github/workflows/linters.yml#L17

glimchb (approved):

Issue #143, opened 2022-09-29
storage: fix objects handles (user or server) see `User-specified IDs`


glimchb:
In the current OPI storage API, the handles are int64 "id"s, and are "mostly" assigned by the gRPC client (except for NVMeSubsystemCreate). 

In the proposed PR #133 handles are keys defined to be 128bits opaque values?

To be most flexible, shouldn't the handles be opaque strings returned by the gRPC server ?

we also don't conform to [Google's AIP](https://google.aip.dev/133) (ex. Create functions should return the created object and not a CreateResponse).

glimchb:
@llabordehpe FYI

jainvipin:
I agree it should be opaque set of `bytes` (not perhaps `strings` that are null terminated). So #133 has it as bytes. Having said that bytes without a length will still require an understanding on the size of it. So 128bit i.e. 16bytes looked good, which can also cover standard UUID in them.

llabordehpe:
If the API provider assigns the handle upon object creation, the API user doesn't need to know the length, right?

jainvipin:
yes @llabordehpe - wouldn't client need to know how many bytes to parse/read, so either we make it null terminated string e.g. uuid that is null terminated, or specify the length somehow. So in that sense I am good with a string either.

jainvipin:
having said that the @glimchb - the way it is defined in #133, it is an opaque string. 128bits is just specified the documentation.

glimchb:
@PWAlessi FYI

glimchb:
@jainvipin @llabordehpe do you agree ? @PWAlessi is implementing this as string...

llabordehpe:
@PWAlessi: I thought there were issues when using "string" for opaque handles. Can you really have any binary content in a tring, like 0-value bytes ?

seroyer:
@llabordehpe no it's not a good idea to directly put binary data into a string value.  However, the implementation that requires a binary value can encode or more generally serialize that value into a string however it likes.  Common choices might be hexadecimal or uuencode.  It's up to the implementation to then take that opaque string and decode/deserialize that value back into whatever it needs on the back end.

glimchb:
all please review #150 for new object type we will use as handle returned by Create()
For now it is string, we can change later on as this conversation in this thread evolves...

jainvipin:
`keyId` string is what we used in updated PRs. should we close this issue?

glimchb:
all the work was completed
[...]

PR #146, opened 2022-10-05
Move Uuid message definition up to top level common directory

PWAlessi:
This is the first step in using Uuids. I moved the Uuid message definition up to a top level common directory because it is used across protos in security and storage.


glimchb:
@jainvipin @llabordehpe FYI

glimchb (approved):
make sense

common/v1/uuid.proto
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2022 Dell Inc, or its subsidiaries.
+// {C} Copyright 2022 Pensando Systems Inc. All rights reserved
+
+syntax = "proto3";
+package opi_api.common.v1;
+option go_package = "github.com/opiproject/opi-api/common/v1";
+
+message Uuid {
+  string value = 1;
```
mestery, line 10: I hate this being a string, and I know @seroyer mentioned this on another PR as well.

glimchb: @mestery we can change this as part of another PR
this one is just moving it to a common place

mestery (commented):
Looks fine.

common/v1/uuid.proto
llabordehpe: If you are trying to define a common message to hold a generic and opaque object handle, I would not call it uuid, since uuid has a special meaning (with layout, length, etc). Wht not call it a handle, a reference, a key, an identifier... whatever but not a uuid.

llabordehpe:
@glimchb : there is no point in asking people's opinion if you merge things with 5 minutes...
As I said, using the name uuid for this common object handle is bad.

glimchb:
> @glimchb : there is no point in asking people's opinion if you merge things with 5 minutes... As I said, using the name uuid for this common object handle is bad.

@llabordehpe This PR is not deciding on the definition of the object handle. We are still discussing with you and here #143 
This is only moving already existing UUIDs to a common place. Since it is just moving code around I merged it, but you are right, I will give couple of days next time to review

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -64,7 +65,7 @@ message NVMeNamespace {
     int64 num_blocks = 8;
     string nguid = 9;
     string eui64 = 10;
-    string uuid = 11;
+    common.v1.Uuid uuid = 11;
```
llabordehpe, line 68: This should not be using the common object handle type (that you unfortunately want to call a uuid..).
This is a TRUE uuid, as defined in the NVMe spec.

glimchb: Yes, this is true UUID defined in NVMe spec
this is NOT an object handle

PR #149, opened 2022-10-05
ci: fix actions after folders restructure

glimchb:


mestery (approved):
thanks for fixing this!

PR #150, opened 2022-10-05
Added ObjectKey type

PWAlessi:
Added ObjectKey type to common protos


glimchb:
@jainvipin @llabordehpe FYI

llabordehpe:
For this ObjectKey message, I would remove all mention of its length in the comment ("128bit (16byte)")...
The comment also indicates that the ObjectKey value is chosen by the API user (i.e. the client). it is not clear to me that it is what we want. I think it should be allocated by the API provider (i.e. the server).

glimchb:
> 

I agree with @llabordehpe

mestery:
> > For this ObjectKey message, I would remove all mention of its length in the comment ("128bit (16byte)")... The comment also indicates that the ObjectKey value is chosen by the API user (i.e. the client). it is not clear to me that it is what we want. I think it should be allocated by the API provider (i.e. the server).
> 
> I agree with @llabordehpe

Piling on here, but I also agree this should be allocated by the server and not the client.

glimchb (approved):
thanks for fixing @PWAlessi

PR #151, opened 2022-10-06
Proposed changes to frontend_nvme_pcie proto.

PWAlessi:
Added handles to object messages
Changed rpc return values to return objects
Renamed Stats and List rpcs to keep returned object names consistent.

### DIFF
--- a/storage/v1/frontend_nvme_pcie.proto
+++ b/storage/v1/frontend_nvme_pcie.proto
@@ -45,24 +45,30 @@ message NVMeSubsystem {
     string serial_number = 2;
     string model_number = 3;
     int64 max_ns = 4;
+
+    // handle is an opaque object handle that is not user settable.
+    // handle will be returned with created object
     opi_api.common.v1.ObjectKey handle = 5;
 }
 
 message NVMeController {
     int64 id = 1;
     string name = 2;
-    string subsystem_id = 3;
+    common.v1.ObjectKey subsystem_handle = 3;
     NvmeControllerPciId pcie_id = 4;
     int64 max_io_qps = 5;
     int64 max_ns = 6;
+
+    // handle is an opaque object handle that is not user settable.
+    // handle will be returned with created object
     common.v1.ObjectKey handle = 7;
 }
 
 message NVMeNamespace {
     int64 id = 1;
     string name = 2;
-    string subsystem_id = 3;
-    int64 controller_id = 4;
+    common.v1.ObjectKey subsystem_handle = 3;
+    common.v1.ObjectKey controller_handle = 4;
     int64 nsid = 5;
     string bdev = 6;
     int64 block_size = 7;
@@ -72,6 +78,9 @@ message NVMeNamespace {
     common.v1.Uuid uuid = 11;
     string multipath = 12;
     string authentication = 13;
+
+    // handle is an opaque object handle that is not user settable.
+    // handle will be returned with created object
     common.v1.ObjectKey handle = 14;
 }
 
@@ -104,7 +113,8 @@ message NVMeSubsystemStatsRequest {
 }
 
 message NVMeSubsystemStats {
-    string stats = 1;
+    common.v1.ObjectKey subsystem_handle = 1;
+    string stats = 2;
 }
 
 message NVMeControllerCreateRequest {
@@ -142,7 +152,7 @@ message NVMeControllerStatsRequest {
 }
 
 message NVMeControllerStats {
-    int64 id = 1;
+    common.v1.ObjectKey controller_handle = 1;
     string stats = 2;
 }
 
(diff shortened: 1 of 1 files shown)


PWAlessi:
This is a draft PR of the proposed changes to the frontend_nvme_pcie API. After discussion, I'll implement changes across the storage APIs.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -174,54 +150,38 @@ message NVMeNamespaceCreateRequest {
...
-    int64 controller_id = 2;
-    int64 namespace_id = 3;
+    common.v1.ObjectKey subsystem_handle = 1;
+    common.v1.ObjectKey controller_handle = 2;
+    common.v1.ObjectKey namespace_handle = 3;
 }
 
-message NVMeNamespaceStatsResponse {
+message NVMeNamespaceStats {
     int64 id = 1;
```
glimchb, line 185: id here should be ObjectKey ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -68,104 +72,76 @@ message NVMeNamespace {
     common.v1.Uuid uuid = 11;
     string multipath = 12;
     string authentication = 13;
+    common.v1.ObjectKey handle = 14;
```
glimchb, line 75: same as commant above 
```
    string subsystem_id = 3;
    int64 controller_id = 4;
```
should be handles ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -6,43 +6,46 @@ package opi_api.storage.v1;
...
+    rpc NVMeNamespaceGet    (NVMeNamespaceGetRequest)    returns (NVMeNamespace)    {}
+    rpc NVMeNamespaceGetStats  (NVMeNamespaceStatsRequest)  returns (NVMeNamespaceStats)  {}
 }
 
 message NVMeSubsystem {
     string nqn = 1;
     string serial_number = 2;
     string model_number = 3;
     int64 max_ns = 4;
+    opi_api.common.v1.ObjectKey handle = 5;
```
glimchb, line 48: by doing this, you force grpc client to create a handle ? 
I was thinking like Louis maybe that server creates handle and returns it in the response to the client... 
WDYT ?

PWAlessi: I wanted to use the same object for the request and response. I figured that requestors would pass an empty handle and the response would contain a populated handle.

glimchb: Ah, ok... let's add comment in the code as well

storage/v1/frontend_nvme_pcie.proto
PWAlessi: ok

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -52,6 +55,7 @@ message NVMeController {
     NvmeControllerPciId pcie_id = 4;
     int64 max_io_qps = 5;
     int64 max_ns = 6;
+    common.v1.ObjectKey handle = 7;
```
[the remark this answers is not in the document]
PWAlessi: 👍

storage/v1/frontend_nvme_pcie.proto
PWAlessi: 👍

storage/v1/frontend_nvme_pcie.proto
llabordehpe: Couldn't we call the ObjectKey field "key" (or name the common type "ObjectHandle") for consistency's sake?
And I would propose to have it as the first field of each object.

PWAlessi: > Couldn't we call the ObjectKey field "key" (or name the common type "ObjectHandle") for consistency's sake? And I would propose to have it as the first field of each object.

Yes, we're early enough on where I can re-order the fields. I didn't do it originally because it's generally not done, but, here, I think it's the right thing to do.

Regarding naming, I think that you've got a good idea. I think that even `ObjectHandle` could be reduced to `Handle` as all handles reference objects.

glimchb: +1

glimchb:
@yadong2007 @prasunkapoor @GottliebNoam @sburla-marvell fyi

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -6,59 +6,69 @@ package opi_api.storage.v1;
...
 
 message NVMeNamespace {
     int64 id = 1;
     string name = 2;
-    string subsystem_id = 3;
-    int64 controller_id = 4;
+    common.v1.ObjectKey subsystem_handle = 3;
+    common.v1.ObjectKey controller_handle = 4;
     int64 nsid = 5;
     string bdev = 6;
```
llabordehpe, line 73: Shouldn't this be a handle to a backend object ?

glimchb: Yes, you need 3 handles here: to subsystem, to controller and to a backend object

PR #154, opened 2022-10-11
Correct go_package path

PWAlessi:
Correct go_package path to new path for generated go files


mestery (approved):
Thanks for fixing this! Will this fix the issues in the `opi-api` repository?

PWAlessi:
It should fix the broken references to the go files. I tested it against my forks and it looked good.

mestery:
> It should fix the broken references to the go files. I tested it against my forks and it looked good.

Sweet! Merged it, lets see if it fixes things now! Thanks again!

PR #156, opened 2022-10-11
firewall: Remove firewall API

mestery:
We have forked the sessionoffload API here [1] and are iterating on it there. We should remove it from the opi-api repository at this point.

[1] https://github.com/opiproject/sessionOffload


glimchb (approved):
YES!

PR #162, opened 2022-10-17
Added AioControllerGetByName to allow retrieval of Aio Controller by

PWAlessi:
Added AioControllerGetByName to allow retrieval of Aio Controller by name

### DIFF
--- a/storage/v1/backend_aio.proto
+++ b/storage/v1/backend_aio.proto
@@ -15,7 +15,6 @@ service AioControllerService {
     rpc AioControllerCreate (AioControllerCreateRequest) returns (AioController) {}
     rpc AioControllerDelete (AioControllerDeleteRequest) returns (google.protobuf.Empty) {}
     rpc AioControllerGet    (AioControllerGetRequest)    returns (AioController)    {}
-    rpc AioControllerGetByName    (AioControllerGetByNameRequest)    returns (AioController)    {}
     rpc AioControllerGetList   (AioControllerGetListRequest)   returns (AioControllerList)   {}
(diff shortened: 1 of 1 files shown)


storage/v1/backend_aio.proto
```diff
@@ -15,6 +15,7 @@ service AioControllerService {
     rpc AioControllerCreate (AioControllerCreateRequest) returns (AioController) {}
     rpc AioControllerDelete (AioControllerDeleteRequest) returns (google.protobuf.Empty) {}
     rpc AioControllerGet    (AioControllerGetRequest)    returns (AioController)    {}
+    rpc AioControllerGetByName    (AioControllerGetByNameRequest)    returns (AioController)    {}
```
glimchb, line 18: how `AioControllerGet` and `AioControllerGetByName` are different ? 
commit message doesn't explain
also why do we need it ? how one would use it... ?

PWAlessi: I can get rid of it. We can always add it back in later if we want to.

storage/v1/backend_aio.proto
```diff
@@ -30,6 +30,7 @@ message AioController {
     int64 block_size = 4;
     int64 num_blocks = 5;
     common.v1.Uuid uuid =6;
+    string filename = 7;
```
glimchb, line 33: also looks like `int64 id = 2;` is redundant...

PWAlessi: Ok, I'll remove it

glimchb (approved):
lgtm

PR #167, opened 2022-10-18
ci: fail if forget to auto-generate protos

glimchb:


seroyer (approved):
Seems like a good idea

PR #169, opened 2022-10-21
frontend proto changes

jainvipin:
Frontend proto updates to incorporates comments in #133, and merges NvmeSubsystem, NvmeController and NvmeNamespace objects. A few notes
- comments beginning with `//??` are really going to be removed but kept for the context and discussions for now
- have included id as an object key to each object, forward object referencing is so easy with that; this is by far the biggest semantic change in this PR
- rearranged some enums, assuming no production/compliant implementations will run into compatibility issues.
- fields are separated by a newline, but we can remove it if the coding guidelines needs to keep it next to each other


storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // serial number must not exceed 'NSV_CTRLR_SERIAL_NO_LEN' bytes
+    //?? val change
+    string serial_number = 3;
+
+    // model number, must not exceed 'NSV_CTRLR_MODEL_NO_LEN' bytes
+    //?? val change
+    string model_number = 4;
+
+    //?? do we need this in subsystem given that controller object would have it
+    //?? val change
```
mkalderon, line 77: we need this since there  could be several controllers in subsystem.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
-    string nqn = 1;
-    string serial_number = 2;
-    string model_number = 3;
-    int64 max_ns = 4;
+    // object's unique identifier
+    common.v1.Uuid id = 1;
+
+    // NVMe subsystem NQN to which the controller belongs
+    // Refer to the NQN format in the NVMe base specifications, must not
+    // exceed 'NSV_NVME_SUBSYSTEM_NQN_LEN' bytes
```
glimchb, line 64: example `nqn.2014-08.org.nvmexpress.discovery`

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    string model_number = 4;
+
+    //?? do we need this in subsystem given that controller object would have it
+    //?? val change
+    int64 max_ns = 5;
+
+    // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
+    string firmware_revision  = 6;
+
+    // FRU identfier, 16bytes opaque identity for the type of unit
```
glimchb, line 83: per @jkogutINTC comment 
> Please reference to `FGUID` in NVMe specification.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    bytes fru_guid = 7;
 }
 
 message NVMeController {
-    int64 id = 1;
+    // object's unique identifier
+    //?? replaces: int64 id = 1;
+    common.v1.Uuid id = 1;
+
+    //?? is this name needed if uuid is uniquely identifying the controller
```
glimchb, line 92: I don't see use in the `name` field of the controller

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
 }
 
 message NVMeNamespace {
-    int64 id = 1;
+    // namespace's unique key
+    // replaces: int64 id = 1;
+    common.v1.Uuid id = 1;
+
+    //?? do we need this if NvmeNamespace is identified using unique uuid
     string name = 2;
```
glimchb, line 125: I don't see use in the `name` field of the namespace

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    NSV_NVME_SUBSYSTEM_NQN_LEN        = 256;
+}
+
 message NVMeSubsystem {
-    string nqn = 1;
-    string serial_number = 2;
-    string model_number = 3;
-    int64 max_ns = 4;
+    // object's unique identifier
+    common.v1.Uuid id = 1;
```
glimchb, line 60: we wanted to use `common.v1.ObjectKey` instead of `common.v1.Uuid` as opaque object's unique identifier
`common.v1.Uuid` should be used only in places per NVME spec that defines true UUID/NGUID/EU64 for example...

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+
+    // FRU identfier, 16bytes opaque identity for the type of unit
+    bytes fru_guid = 7;
 }
 
 message NVMeController {
-    int64 id = 1;
+    // object's unique identifier
+    //?? replaces: int64 id = 1;
+    common.v1.Uuid id = 1;
```
glimchb, line 90: we wanted to use `common.v1.ObjectKey` instead of `common.v1.Uuid` as opaque object's unique identifier
`common.v1.Uuid` should be used only in places per NVME spec that defines true UUID/NGUID/EU64 for example...

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // the controller.
+    //?? val change
+    uint32 max_ns = 7;
 }
 
 message NVMeNamespace {
-    int64 id = 1;
+    // namespace's unique key
+    // replaces: int64 id = 1;
+    common.v1.Uuid id = 1;
```
glimchb, line 122: we wanted to use `common.v1.ObjectKey` instead of `common.v1.Uuid` as opaque object's unique identifier
`common.v1.Uuid` should be used only in places per NVME spec that defines true UUID/NGUID/EU64 for example...

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
     string nguid = 9;
-    string eui64 = 10;
+
+    // 64bit Extended unique identifier for the namespace
+    // mandatory if guid is not specified, optional otherwise
+    //?? replaces: string eui64 = 10 (fixed64 allows user to specify the byte order)
+    fixed64 eui64 = 10;
+
+    // Globally unique identifier for the namespace
     common.v1.Uuid uuid = 11;
```
glimchb, line 161: here `common.v1.Uuid` is actually correct

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+
+    //?? do we need this if NvmeNamespace is identified using unique uuid
     string name = 2;
-    string subsystem_id = 3;
-    int64 controller_id = 4;
-    int64 nsid = 5;
+
+    //?? can we take this out if controller is referring to a subsystem anyways
+    //?? replaces string subsystem_id = 3;
+    common.v1.Uuid subsystem_id = 3;
```
glimchb, line 129: we wanted to use `common.v1.ObjectKey` instead of `common.v1.Uuid` as opaque object's unique identifier

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+
+    // key of the PCIe controller object that will host this namespace.
+    //?? replaces int64 controller_id = 4;
+    common.v1.Uuid controller_id = 4;
```
glimchb, line 133: we wanted to use `common.v1.ObjectKey` instead of `common.v1.Uuid` as opaque object's unique identifier

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // 64bit Extended unique identifier for the namespace
+    // mandatory if guid is not specified, optional otherwise
+    //?? replaces: string eui64 = 10 (fixed64 allows user to specify the byte order)
+    fixed64 eui64 = 10;
+
+    // Globally unique identifier for the namespace
     common.v1.Uuid uuid = 11;
+
+    //?? what is this for
     string multipath = 12;
```
glimchb, line 164: no use for now, wanted to reserve for future, can remove now after discussion

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+
+    //?? what is this for
     string authentication = 13;
```
glimchb, line 167: no use for now, wanted to reserve for future, can remove now after discussion

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    common.v1.Uuid controller_id = 4;
+
+    // NSID present to the host by the NVMe PCIe controller.
+    // If not provided, then the controller will assign an unused NSID
+    // within the max namespace range - auto assigned nsid may not work
+    // for live migration
+    //?? replaces: int64 nsid = 5;
+    uint32 host_nsid = 5;
+
     string bdev = 6;
```
glimchb, line 142: need to discuss this `bdev` as well... I think should be reference to either middle-end object (like encryption, compression, ...) or to the back-end object, like nvmeof bdev... wdyt ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
     common.v1.Uuid uuid = 11;
+
+    //?? what is this for
     string multipath = 12;
+
+    //?? what is this for
     string authentication = 13;
+
+    // reference to encryption key for the data at rest encryption
+    common.v1.Uuid crypto_key_id = 14;
```
glimchb, line 170: does this belong here ? I think should be reference to either middle-end object (like encryption, compression, ...) or to the back-end object, like nvmeof bdev... wdyt ?

glimchb:
@jkogutINTC please re-review this as well

storage/v1/frontend_nvme_pcie.proto
mkalderon: @glimchb do you mean replace bdev with objectKey ?

glimchb: yeah, maybe... objectKey  that points to a backend device

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // maximum allowed controller model number string length
+    NSV_CTRLR_MODEL_NO_LEN  = 40;
+
+    // maximum allowed firmware revision string length
+    NSV_CTRLR_FW_REV_LEN    = 8;
+
+    // maximum length for NQN field
+    NSV_NVME_SUBSYSTEM_NQN_LEN        = 256;
+}
+
```
benlwalker, line 57: As I've said elsewhere, there's likely no reason to expose a subsystem object in the front-end. An xPU will probably never expose more than 1 controller per subsystem, so the two objects can be grouped into a "device" object. But if someone really has a use case for this in an xPU then I'm listening.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
     rpc NVMeNamespaceStats  (NVMeNamespaceStatsRequest)  returns (NVMeNamespaceStatsResponse)  {}
 }
 
+enum NvmeFrontEndConstants {
```
benlwalker, line 42: Aren't these all spec defined values?

storage/v1/common.proto
```diff
@@ -14,11 +14,26 @@ option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
...
     // Physical function, always set to 0 in current model
     uint32 function         = 3;
+
     // SRIOV Virtual function within the Device and Physical function.
     // Set to 0 for Physical Function. Virtual Function numbering starts from 1
     uint32 virtual_function  = 4;
 }
+
+// Path Management Selection
+enum NvmePathMgmtType {
```
benlwalker, line 30: Can you explain what these do? Multipath is almost certainly a property of the backend and shouldn't be in this PR.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // maximum host IO queue pairs allowed, value will default to
+    // limits in PCI device configuration; if set to 0 or more it
+    // will default to maximum permitted per xPU's capability
+    //?? val change
+    uint32 max_io_qps = 6;
+
+    // maximum Number of namespaces that will be provisioned under
+    // the controller.
+    //?? val change
+    uint32 max_ns = 7;
```
benlwalker, line 116: This does not seem necessary. It's in the subsystem.

storage/v1/frontend_nvme_pcie.proto
glimchb: why do you say that ? I think NS belongs to ctrl not subsystem
see

storage/v1/frontend_nvme_pcie.proto
glimchb: NS is attached to 1 or more controllers

storage/v1/frontend_nvme_pcie.proto
glimchb: yep, we should use https://github.com/opiproject/opi-api/blob/main/common/v1/object_key.proto

storage/v1/frontend_nvme_pcie.proto
glimchb: - backend is we talking about PSK's for TLS for example
- middleend is we talking about data at rest

storage/v1/frontend_nvme_pcie.proto
benlwalker: A namespace gets assigned to a subsystem in NVMe, not a controller. This is directly from the NVMe spec and that's how your diagram shows it too. The namespaces are in the orange box, not in the blue boxes.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // If not provided, then the controller will assign an unused NSID
+    // within the max namespace range - auto assigned nsid may not work
+    // for live migration
+    //?? replaces: int64 nsid = 5;
+    uint32 host_nsid = 5;
+
     string bdev = 6;
+
+    // Block size in bytes, must be power of 2 and must be less than the max
+    // io size supported. Typically tested values are 512, and 4k.
```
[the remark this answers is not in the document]
glimchb: if those are optional fields that if left empty - taken from backend, I don;t mind leaving them for now here and re-cosider again once we defined middle end. baby steps

storage/v1/frontend_nvme_pcie.proto
benlwalker: No. A namespace is placed in a subsystem.

storage/v1/frontend_nvme_pcie.proto
glimchb: yep, what do you suggest ? 
we can define them like this https://github.com/linux-nvme/libnvme/blob/master/src/nvme/types.h#L99-L129

storage/v1/frontend_nvme_pcie.proto
glimchb: we see use cases with more then 1 controller per subsystem

storage/v1/frontend_nvme_pcie.proto
glimchb: so ctrl only have NSID inside ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+    // key of the PCIe controller object that will host this namespace.
+    //?? replaces int64 controller_id = 4;
+    common.v1.Uuid controller_id = 4;
+
+    // NSID present to the host by the NVMe PCIe controller.
+    // If not provided, then the controller will assign an unused NSID
+    // within the max namespace range - auto assigned nsid may not work
+    // for live migration
+    //?? replaces: int64 nsid = 5;
+    uint32 host_nsid = 5;
```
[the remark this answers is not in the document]
glimchb: should `host_nsid` belong to controller instead then ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
     string authentication = 13;
+
+    // reference to encryption key for the data at rest encryption
+    common.v1.Uuid crypto_key_id = 14;
+
+    // optimal write size hint to host driver. Host IO stack may use
+    // this to regulate IO size. Must be a multiple of the preferred write
+    // granularity. Must not exceed the controller maximum IO size value
+    // configured in the nvme agent config file.
+    uint32 optimal_write_size = 15;
```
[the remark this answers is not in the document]
glimchb: I think the intent here is what you present to the HOST over PCIe

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,36 +39,146 @@ service NVMeNamespaceService {
...
+
+    // preferred write granularity hint to the host driver. Host IO
+    // stack may use this to align IO sizes to the write granularity for
+    // optimum performance.
+    uint32 pref_write_granularity= 16;
```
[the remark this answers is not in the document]
glimchb: I think the intent here is what you present to the HOST over PCIe

storage/v1/frontend_nvme_pcie.proto
benlwalker: I suggest dropping this enum since we know all of the values by spec.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: It's not for the front-end, or even an administrator, to decide. It's based on the BE/ME device reporting this value.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: Emulation is often much trickier than it seems, so my suggested baby step is just delete these two things for now.

storage/v1/frontend_nvme_pcie.proto
benlwalker: Unfortunately, that's complicated. There's actually two types of NSID - private and public. Public NSIDs are inside the subsystem, and all controllers see the same set. Private NSIDs are inside the controller. You can think of a private NSID as an extra layer of indirection that aliases the public NSID to a private value, although strictly speaking the private NSID does not have to also appear in the public list.

I've never seen any implementation ever use a private NSID. Certainly, lots of implementations change which public NSIDs are visible to hosts based on access control, but I've never seen an implementation use a private NSID. I don't believe they actually add any functionality, security, or value generally. I do not think we need to support private NSIDs in OPI.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: The correct model is that there's an map structure inside the subsystem, whose key is NSID and whose value is the namespace object. The namespace doesn't have a back-reference, nor does it care what it's NSID is. I suggested moving this to the top level RPC that is adding this namespace to the subsystem. With dispersed namespaces, or private namespaces, you could end up adding this namespace to multiple different things at different NSIDs.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
glimchb: but what if implementation inside DPU/IPU for NVMe emulation has constrains of it's own on top of BE/ME device reporting ?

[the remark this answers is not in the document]
benlwalker: Then those constraints do not need to be configured over OPI and can just be set up and reported by the device

storage/v1/frontend_nvme_pcie.proto
jainvipin: yes, that should serve generically as well. So, I am good with an opaque string instead of uuid. let's hear from others if anyone wants to opine on it.

storage/v1/frontend_nvme_pcie.proto
jainvipin: yes will change.

storage/v1/frontend_nvme_pcie.proto
jainvipin: will change.

storage/v1/frontend_nvme_pcie.proto
jainvipin: ack

storage/v1/frontend_nvme_pcie.proto
jainvipin: yes I am not sure about the use for it. if not we can remove/move it to the backend objects.

storage/v1/frontend_nvme_pcie.proto
jainvipin: well, ultimately encryption granularity will need to be specified somewhere, so I put this in here. We can move this to ObjectKey instead, like in other objects; so this is really an attachment point for crypto, not the middleend layer, but a glue.

storage/v1/frontend_nvme_pcie.proto
jainvipin: ack

storage/v1/frontend_nvme_pcie.proto
jainvipin: @glimchb - I will include the example in the comments above

storage/v1/frontend_nvme_pcie.proto
jainvipin: agree @benlwalker  - but I kept it since this was already an object and so moved a few things in there. So I am good removing this all together from the front end objects - however some of the fields included in Subsystem object can be folded into the NvmeController object then.

storage/v1/frontend_nvme_pcie.proto
jainvipin: I am not sure if they are, but even if they do, putting the here may serve explicitly for different current/future versions.

storage/v1/common.proto
jainvipin: yes multipathing is a backend property (separate PR). however this is allowing selecting the path to be used by a given frontend Namespace object. So think of this as an association to a backend definition.

storage/v1/frontend_nvme_pcie.proto
benlwalker: +1 from me. I'm not entirely convinced the server shouldn't be the one generating the handles, but I'm fine with moving forward in this direction for now.

storage/v1/frontend_nvme_pcie.proto
jainvipin: I was thinking we keep it here  and take it out of the subsystem; it also depends on if we take out the subsystem object all together.

storage/v1/frontend_nvme_pcie.proto
jainvipin: thanks for catching - @glimchb, I will make this change in subsequent commit.

storage/v1/frontend_nvme_pcie.proto
jainvipin: yes a commit on this PR coming to move these all keys to reflect that.

storage/v1/frontend_nvme_pcie.proto
jainvipin: ack

storage/v1/frontend_nvme_pcie.proto
jainvipin: fixed in recent commit - all uuids are now objectKey.

storage/v1/frontend_nvme_pcie.proto
jainvipin: > I suggest dropping this enum since we know all of the values by spec.

specifications also have versions, having this allows code to work on it via APIs, isn't it? Since these fields are exchanged via a versioned APIs, wouldn't it be better to keep it as part of API, instead of defining this in both client and server. Ultimately the client and server would need to have some checks on the length.

storage/v1/frontend_nvme_pcie.proto
jainvipin: fixed in updated commit

storage/v1/frontend_nvme_pcie.proto
benlwalker: Are we expecting a single xPU to support different NVMe versions? Setting the version would be a clearer way to communicate that, rather than these lengths. Or are you suggesting some xPUs have stricter limits on the lengths of these values than the spec does, and this is for the xPU to report that back up to the client?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -38,44 +40,142 @@ service NVMeNamespaceService {
...
-    int64 id = 1;
-    string name = 2;
-    string subsystem_id = 3;
+    // object's unique identifier
+    //?? replaces: int64 id = 1;
+    common.v1.ObjectKey id = 1;
+
+    // subsystem controller id range: 0 to 65535.
+    // must not be reused under the same subsystem
+    uint32 nvme_controller_id  = 2;
```
jainvipin, line 95: @jkogutINTC @glimchb - per the comments in PR #133 I tried changing this to uint16, but proto compiler is throwing an error; so I kept it as uint32 with a comment specifying the range of the values.

storage/v1/frontend_nvme_pcie.proto
glimchb: this brings up a question - do we need to have NMVE spec version xpu supports ? like 1.4 ?

storage/v1/common.proto
benlwalker: I don't think that makes any sense. You create the backend object, which includes the possible paths, and then you assign that object to the front-end. This should be removed from the front-end configuration.

storage/v1/frontend_nvme_pcie.proto
jainvipin: perhaps-  let's do in a separate PR then after more discussions/thought.

glimchb: +1

jainvipin:
push the final set of changes - pls review @benlwalker @glimchb @mkalderon

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -39,47 +41,116 @@ service NVMeNamespaceService {
...
+
+    // 64bit Extended unique identifier for the namespace
+    // mandatory if guid is not specified, optional otherwise
+    fixed64 eui64 = 9;
+
+    // Globally unique identifier for the namespace
+    common.v1.Uuid uuid = 10;
+
+    // reference to encryption key for the data at rest encryption
+    common.v1.ObjectKey crypto_key_id = 11;
```
glimchb, line 130: to be more generic, this can be renamed to backend object id
type is ok here
once renamed, can remove the field `bdev` above

glimchb (commented):
Looks good now from my perspective
One last nit about `crypto_key_id` inline
Not clicking "approve" button only because want to give till the end of the week for other people to comment and if no objection, will merge on Thursday

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -144,29 +214,27 @@ message NVMeControllerUpdateResponse {
...
 
 message NVMeControllerGetRequest {
-    int64 subsystem_id = 1;
-    int64 controller_id = 2;
+    common.v1.ObjectKey controller_id = 1;
 }
 
 message NVMeControllerGetResponse {
-    NVMeController controller = 1;
+    NVMeController controller_id = 1;
```
glimchb, line 229: this suppose to retunr the controller object, not just and ID

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -103,15 +174,15 @@ message NVMeSubsystemListResponse {
 }
 
 message NVMeSubsystemGetRequest {
-    string nqn = 1;
+    common.v1.ObjectKey subsystem_id = 1;
```
glimchb, line 177: this suppose to return the subsystem object, not just and ID

storage/v1/frontend_nvme_pcie.proto
jainvipin: sure @glimchb - did you mean middle-end or the backend object-id? What we need to decide is if frontned objects will have reference to a generic bucket of middleend services (e.g. crypto, etc.) or one reference per frontend object say encryption granularity may make sense at a different level than compression or other middle-end services.

storage/v1/frontend_nvme_pcie.proto
jainvipin: yes it refers to object, but renamed it enmass as controller_id; so let me fix this. good catch!

jainvipin: I mean it was returning the object, just the name is incorrectly changed to call it id; pushed the fix.

glimchb: thanks

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
mkalderon: agree with @benlwalker that these should be dropped. If not, then it needs to be clarified to determine with takes precedence - values from backend object or values from the api. Also for scalar types since the ‘ default/value not set’ would be 0 it could also be a valid value, so it's not entirely harmful. I think the comment is also relevant for the fields: optimal_write_size, write_granularity below

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -39,47 +41,114 @@ service NVMeNamespaceService {
...
+    // object's unique identifier
+    common.v1.ObjectKey id = 1;
+
+    // NVMe subsystem NQN to which the controller belongs
+    // Refer to the NQN format in the NVMe base specifications, must not
+    // exceed 'NSV_NVME_SUBSYSTEM_NQN_LEN' bytes
+    string nqn = 2;
+
+    // serial number must not exceed 'NSV_CTRLR_SERIAL_NO_LEN' bytes
+    string serial_number = 3;
```
mkalderon, line 53: is NSV_CTRLR_SERIAL_NO_LEN etc.. defined in the end somewhere?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -39,47 +41,114 @@ service NVMeNamespaceService {
...
+    string model_number = 4;
+
+    // maximum namespaces within a subsystem
+    int64 max_ns = 5;
+
+    // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
+    string firmware_revision  = 6;
+
+    // FRU identfier, 16bytes opaque identity for the type of unit
+    bytes fru_guid = 7;
```
mkalderon, line 65: @jainvipin aren't the firmware_revision and fru_guid part of the controller properties and not the subsystem ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -39,47 +41,114 @@ service NVMeNamespaceService {
...
+
+    // xPU's PCI ID for the controller
     NvmeControllerPciId pcie_id = 4;
-    int64 max_io_qps = 5;
-    int64 max_ns = 6;
+
+    // maximum host IO queue pairs allowed, value will default to
+    // limits in PCI device configuration; if set to 0 or more it
+    // will default to maximum permitted per xPU's capability
+    uint32 max_io_qps = 5;
```
mkalderon, line 85: this limits the ability to define seperate CQ, SQ and attach several SQs to a single CQs as spec permits.
Perhaps should be split ? 
also - perhaps we should add the max_queue_depth param ?

storage/v1/frontend_nvme_pcie.proto
glimchb: we postponed `NSV_CTRLR_SERIAL_NO_LEN` to next PR

storage/v1/frontend_nvme_pcie.proto
glimchb: I suggest if you can please create new PR @mkalderon with both additions:
- split CQ and SQ
- add `max_queue_depth`

storage/v1/frontend_nvme_pcie.proto
mkalderon: so should it still be part of comment ? not critical.. just made me look for it :-)

storage/v1/frontend_nvme_pcie.proto
mkalderon: ok

storage/v1/frontend_nvme_pcie.proto
glimchb: ah, ok... yes, need to remove the comment

glimchb:
Approved during the API meeting nobody objected. Merging now

glimchb (approved):

Issue #142, opened 2022-09-29
add or remove auto-gen files


glimchb:
see #139 vs #141

PWAlessi:
FYI, Google just provides the .proto files for their APIs (see https://github.com/googleapis/googleapis/tree/master/google).  Directory structure typically ./service/v1/*.proto (ex. google ai platform https://github.com/googleapis/googleapis/tree/master/google/cloud/aiplatform/v1)

Should we move to this structure? I believe that we only want to provide the .proto files as they are the interface definitions.

jainvipin:
auto-gen files create a nuisance when merging, but helps avoid additional code generation by the users of the APIs.

I would prefer keeping generated files and language bindings/packages so they can act as reference and also avoid burden on the users of the APIs (which would be inherently in a specific language). Of course developers have to deal with merge conflicts resolution on generated files, which can be handled easily (although unwanted)

glimchb:
what we landing on now is separate repo for only generated files....

every time PR for proto is submitted, CI/CD github action will auto-create and commit generated files
and even maybe bump a version

jainvipin:
makes sense. thanks! good idea.

PR #170, opened 2022-10-21
[storage] Makefile improvements

benlwalker:
* Remove all generated code from the storage part of the repo. It's generated, so it does not need to be version controlled.
* Simplify the Makefile by removing docker (protoc is packaged on most distros so it's trivial to install) and adding a convenience function.


glimchb (commented):
Great, thanks. This address #142 
 
But without generated code golang failed for me in opi-spdk-bridge repo…. Do we need somehow to generate the code there ?

also why remove the v1 ? This is proto versioning….

benlwalker:
I might need to have it generate code. Where exactly are you seeing an error? When building opi-spdk-bridge?

Also I completely wrecked the existing storage github action and it needs to be rewritten another way. I'll work on that.

glimchb:
> I might need to have it generate code. Where exactly are you seeing an error? When building opi-spdk-bridge?

Yes, when running go get and go build i see failure.

> Also I completely wrecked the existing storage github action and it needs to be rewritten another way. I'll work on that.

Sure, go ahead

glimchb:
We also need to keep either v1 or v1alpha versioning somehow

benlwalker:
I can easily keep the v1 directory, but I did not understand what purpose that served in the repo itself.

glimchb:
> I can easily keep the v1 directory, but I did not understand what purpose that served in the repo itself.

Like this example we have folders every time new api version released https://github.com/tektoncd/pipeline/tree/main/pkg/apis/pipeline

At least this is my groc versioning understanding

PWAlessi:
v1 directory serves to version the APIs. Take a look at [Uber's Protobuf Style Guide V2](https://github.com/uber/prototool/blob/dev/style/README.md#package-versioning) and [Google's API directory structure](https://github.com/googleapis/googleapis/tree/master/google/bigtable). 

Versioning and serving the generated artifacts makes it easier for Go developers to get versioned artifacts via `go get`. It's like publishing jars to artifactory.

glimchb (changes requested):
v1 folder and golang generations have to be resolved before merging this

benlwalker:
@glimchb After review of golang requirements and practices, I think the best approach is to have a dedicated, separate repository for the go bindings, and then sync to that using a github action from this repository. Are you able to create the go bindings repo? Do we need to discuss this somewhere and get everyone to agree? I can present pros and cons.

glimchb:
no need to discuss, @benlwalker I agree
The committed generated proto files with sources in the same repo was a temp hack until we ramp up

benlwalker:
I've reduced this PR just to the makefile cleanups. I'll do a separate PR for deleting the generated code once we have the dedicated repo set up.

glimchb (changes requested):
missing `google-api-linter`

benlwalker:
Did we decide to make a new repo to hold the golang bindings? What's the conclusion there?

glimchb (commented):
looks better, thanks @benlwalker

jainvipin:
> @glimchb After review of golang requirements and practices, I think the best approach is to have a dedicated, separate repository for the go bindings, and then sync to that using a github action from this repository. Are you able to create the go bindings repo? Do we need to discuss this somewhere and get everyone to agree? I can present pros and cons.

having client bindings in a separate repo would be good.

jainvipin (approved):
LGTM

intelfisz (approved):
Are we waiting for GH actions adjustments to this PR yet or can it already be merged?

storage/Makefile
```diff
@@ -2,19 +2,89 @@
 # Copyright (C) 2022 Intel Corporation
 # Copyright (c) 2022 Dell Inc, or its subsidiaries.
 
-all:
-	rm -rf ./v1alpha1/{autogen.md,gen,google}
-	mkdir -p  ./v1alpha1/gen/{go,cpp,python,java}
+CURRENT_VERSION := v1alpha1
+COMMON_VERSION := v1
```
jainvipin, line 6: do we want to call it `v1` as yet?

jainvipin (commented):
this is a good positive change - only one comment, otherwise looks good to me. If this works for storage, we can adopt the same for networking as well.

sandersms:
We have transitioned to using "buf" for generation of the protobuf bindings which are included in the makefiles.  The plan is to move the bindings to a separate repo which is still being worked.  Versioning of the various APIs needs to also be addressed in the current changes.

The changes here need to be reviewed/updated for inclusion as needed as some of the changes are valuable.

PR #171, opened 2022-10-24
storage: Unify front-end block APIs

benlwalker:
This is a new proposed starting point for a unified set of front-end block APIs can that can work for virtio-blk, NVMe, and NVMe-oF. To be clear right up front, I'm not even certain that I agree with all of these decisions or approve of this set of APIs. But I think it's a good starting point.

What I tried to accomplish here was to unify the front-end APIs but keep the front-end/middle-end/back-end concept in the current OPI model. I've personally been arguing that we need to instead batch these APIs into fewer endpoints for atomicity reasons, but I think at least conceptually the front/middle/back concept is here to stay.

The most important concepts here are as follows:

1) I'm unifying not only exposing an NVMe or virtio-blk device to a host system, but also exposing an NVMe-oF device over the network (so the network is the front end bus).

2) I've introduced a generic "Device" which for NVMe is equivalent to a subsystem.

3) I've introduced the concept of an endpoint. For NVMe/PCIe, this is a controller at some BDF. For other types, it can be re-used and the controller is a more dynamic and short-lived thing.

4) The PCI stuff does not allow the endpoint to specify the BDF. It only lets you pick the PF/VF ids. The BDF is assigned by the OS PCI driver and we cannot pick it in these APIs.

### DIFF
--- a/storage/v1/common.proto
+++ b/storage/v1/common.proto
@@ -6,22 +6,13 @@ syntax = "proto3";
 package opi_api.storage.v1;
 option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
 
-// The controller PCI-ID is used to address a given virtual controller. Virtual
-// controllers are organized into devices with Physical functions and SRIOV
-// virtual function within the physical functions. Currently, xPUs may
-// expose multiple devices with one physical function each and one or more
-// virtual functions under the physical function.
-message NvmeControllerPciId {
-    // Bus number, provided for future usage if needed. Currently set to ’0’
-    uint32 bus              = 1;
-
-    // Device number, based on the NVMe device layout
-    uint32 device           = 2;
-
-    // Physical function, always set to 0 in current model
-    uint32 function         = 3;
-
-    // SRIOV Virtual function within the Device and Physical function.
-    // Set to 0 for Physical Function. Virtual Function numbering starts from 1
-    uint32 virtual_function  = 4;
+// Many front-ends expose PCI devices to the host. This represents that endpoint.
+// The Bus:Device:Function numbers will be assigned by the OS PCI driver, so we
+// cannot set those directly here. What we can pick is which physical function
+// and virtual function the xPU uses.
+message PciEndpoint {
+    // Physical function index
+    uint32 physical_id = 1;
+    // Virtual function index
+    uint32 virtual_id = 2;
 }
--- a/storage/v1/frontend_nvme_pcie.proto
+++ b/storage/v1/frontend_nvme_pcie.proto
@@ -40,25 +40,14 @@ message NVMeSubsystem {
     // object's unique identifier
     common.v1.ObjectKey id = 1;
 
-    // NVMe subsystem NQN to which the controller belongs
-    // Refer to the NQN format in the NVMe base specifications, must not
-    // exceed 'NSV_NVME_SUBSYSTEM_NQN_LEN' bytes
-    string nqn = 2;
-
     // serial number must not exceed 'NSV_CTRLR_SERIAL_NO_LEN' bytes
-    string serial_number = 3;
+    string serial_number = 2;
 
     // model number, must not exceed 'NSV_CTRLR_MODEL_NO_LEN' bytes
-    string model_number = 4;
+    string model_number = 3;
 
     // maximum namespaces within a subsystem
-    int64 max_ns = 5;
-
-    // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
-    string firmware_revision  = 6;
-
-    // FRU identfier, 16bytes opaque identity for the type of unit
-    bytes fru_guid = 7;
+    int32 max_ns = 4;
 }
 
 message NVMeController {
@@ -69,70 +58,35 @@ message NVMeController {
     // must not be reused under the same subsystem
     uint32 nvme_controller_id  = 2;
 
-    // subsystem information
-    common.v1.ObjectKey subsystem_id = 3;
-
     // xPU's PCI ID for the controller
-    PciEndpoint pcie_id = 4;
+    PciEndpoint pcie_id = 3;
 
     // maximum host IO queue pairs allowed, value will default to
     // limits in PCI device configuration; if set to 0 or more it
     // will default to maximum permitted per xPU's capability
-    uint32 max_io_qps = 5;
+    uint32 max_io_qps = 4;
 
     // maximum Number of namespaces that will be provisioned under
     // the controller.
-    uint32 max_ns = 6;
+    uint32 max_ns = 5;
 }
 
 message NVMeNamespace {
     // namespace's unique key
-    // replaces: int64 id = 1;
     common.v1.ObjectKey id = 1;
 
-    // subsystem for this namespace
-    common.v1.ObjectKey subsystem_id = 2;
-
-    // key of the PCIe controller object that will host this namespace.
-    common.v1.ObjectKey controller_id = 3;
-
-    // NSID present to the host by the NVMe PCIe controller.
-    // If not provided, then the controller will assign an unused NSID
-    // within the max namespace range - auto assigned nsid may not work
-    // for live migration
-    uint32 host_nsid = 4;
-
-    // Block size in bytes, must be power of 2 and must be less than the max
-    // io size supported. Typically tested values are 512, and 4k.
-    int64 block_size = 5;
-
-    // Size/Capacity of the namespace in blocks, size in bytes will
-    // be BlockSize x NumBlocks.
-    int64 num_blocks = 6;
-
     // Globally unique identifier for the namespace
-    string nguid = 7;
+    string nguid = 2;
 
     // 64bit Extended unique identifier for the namespace
     // mandatory if guid is not specified, optional otherwise
-    fixed64 eui64 = 8;
+    fixed64 eui64 = 3;
 
     // Globally unique identifier for the namespace
-    common.v1.Uuid uuid = 9;
+    common.v1.Uuid uuid = 4;
 
-    // reference to encryption key for the data at rest encryption
-    common.v1.ObjectKey crypto_key_id = 10;
-
-    // optimal write size hint to host driver. Host IO stack may use
-    // this to regulate IO size. Must be a multiple of the preferred write
-    // granularity. Must not exceed the controller maximum IO size value
-    // configured in the nvme agent config file.
-    uint32 optimal_write_size = 11;
-
-    // preferred write granularity hint to the host driver. Host IO
-    // stack may use this to align IO sizes to the write granularity for
-    // optimum performance.
-    uint32 pref_write_granularity= 12;
+    // The back/middle-end volume to back this namespace.
+    common.v1.ObjectKey volume_id = 5;
 }
 
 message NVMeSubsystemCreateRequest {
@@ -185,6 +139,9 @@ message NVMeSubsystemStatsResponse {
 
 message NVMeControllerCreateRequest {
     NVMeController controller = 1;
+
+    // The subsystem to associate this controller with
+    common.v1.ObjectKey subsystem_id = 2;
 }
 
 message NVMeControllerCreateResponse {
@@ -234,6 +191,12 @@ message NVMeControllerStatsResponse {
 
 message NVMeNamespaceCreateRequest {
     NVMeNamespace namespace = 1;
+
+    common.v1.ObjectKey subsystem_id = 2;
+
+    // The namespace ID inside of the subsystem where this namespace
+    // can be found.
+    uint32 nsid = 3;
 }
 
 message NVMeNamespaceCreateResponse {
@@ -257,8 +220,11 @@ message NVMeNamespaceUpdateResponse {
 }
 
 message NVMeNamespaceListRequest {
-    common.v1.ObjectKey subsystem_id = 1;
-    common.v1.ObjectKey controller_id = 2;
+    oneof view {
+        common.v1.ObjectKey subsystem_id = 1;
+
+        common.v1.ObjectKey controller_id = 2;
+    }
 }
 
 message NVMeNamespaceListResponse {
--- a/storage/v1/frontend_nvme_pcie.proto
+++ b/storage/v1/frontend_nvme_pcie.proto
@@ -13,25 +13,21 @@ import "object_key.proto";
 
 // NVMe/PCIe emulation
 
-service NVMeSubsystemService {
+service FrontendNvme {
     rpc NVMeSubsystemCreate (NVMeSubsystemCreateRequest) returns (NVMeSubsystemCreateResponse) {}
     rpc NVMeSubsystemDelete (NVMeSubsystemDeleteRequest) returns (NVMeSubsystemDeleteResponse) {}
     rpc NVMeSubsystemUpdate (NVMeSubsystemUpdateRequest) returns (NVMeSubsystemUpdateResponse) {}
     rpc NVMeSubsystemList   (NVMeSubsystemListRequest)   returns (NVMeSubsystemListResponse)   {}
     rpc NVMeSubsystemGet    (NVMeSubsystemGetRequest)    returns (NVMeSubsystemGetResponse)    {}
     rpc NVMeSubsystemStats  (NVMeSubsystemStatsRequest)  returns (NVMeSubsystemStatsResponse)  {}
-}
 
-service NVMeControllerService {
     rpc NVMeControllerCreate (NVMeControllerCreateRequest) returns (NVMeControllerCreateResponse) {}
     rpc NVMeControllerDelete (NVMeControllerDeleteRequest) returns (NVMeControllerDeleteResponse) {}
     rpc NVMeControllerUpdate (NVMeControllerUpdateRequest) returns (NVMeControllerUpdateResponse) {}
     rpc NVMeControllerList   (NVMeControllerListRequest)   returns (NVMeControllerListResponse)   {}
     rpc NVMeControllerGet    (NVMeControllerGetRequest)    returns (NVMeControllerGetResponse)    {}
     rpc NVMeControllerStats  (NVMeControllerStatsRequest)  returns (NVMeControllerStatsResponse)  {}
-}
 
-service NVMeNamespaceService {
     rpc NVMeNamespaceCreate (NVMeNamespaceCreateRequest) returns (NVMeNamespaceCreateResponse) {}
     rpc NVMeNamespaceDelete (NVMeNamespaceDeleteRequest) returns (NVMeNamespaceDeleteResponse) {}
     rpc NVMeNamespaceUpdate (NVMeNamespaceUpdateRequest) returns (NVMeNamespaceUpdateResponse) {}
--- a/storage/v1/frontend_nvme_pcie.proto
+++ b/storage/v1/frontend_nvme_pcie.proto
@@ -77,7 +77,7 @@ message NVMeController {
     common.v1.ObjectKey subsystem_id = 3;
 
     // xPU's PCI ID for the controller
-    NvmeControllerPciId pcie_id = 4;
+    PciEndpoint pcie_id = 4;
 
     // maximum host IO queue pairs allowed, value will default to
     // limits in PCI device configuration; if set to 0 or more it
--- a/storage/v1/frontend_virtio_blk.proto
+++ b/storage/v1/frontend_virtio_blk.proto
@@ -22,7 +22,7 @@ service VirtioBlkService {
 message VirtioBlk {
     int64 id = 1;
     string name = 2;
-    NvmeControllerPciId pcie_id = 3;
+    PciEndpoint pcie_id = 3;
     string bdev = 4;
     int64 max_io_qps = 5;
     string serial_number = 6;
--- a/storage/v1/frontend_virtio_blk.proto
+++ b/storage/v1/frontend_virtio_blk.proto
@@ -5,12 +5,13 @@ syntax = "proto3";
 package opi_api.storage.v1;
 option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
 import "common.proto";
+import "object_key.proto";
 
 // Front End (host-facing) APIs.
 
 // Virtio-blk emulation
 
-service VirtioBlkService {
+service FrontendVirtioBlk {
     rpc VirtioBlkCreate (VirtioBlkCreateRequest) returns (VirtioBlkCreateResponse) {}
     rpc VirtioBlkDelete (VirtioBlkDeleteRequest) returns (VirtioBlkDeleteResponse) {}
     rpc VirtioBlkUpdate (VirtioBlkUpdateRequest) returns (VirtioBlkUpdateResponse) {}
@@ -19,17 +20,22 @@ service VirtioBlkService {
     rpc VirtioBlkStats  (VirtioBlkStatsRequest)  returns (VirtioBlkStatsResponse)  {}
 }
 
-message VirtioBlk {
-    int64 id = 1;
-    string name = 2;
-    PciEndpoint pcie_id = 3;
-    string bdev = 4;
-    int64 max_io_qps = 5;
-    string serial_number = 6;
+message VirtioBlkController {
+    common.v1.ObjectKey id = 1;
+
+    // The PCI endpoint where this device should appear
+    PciEndpoint pcie_id = 2;
+
+    // The back/middle-end volume to back this controller.
+    common.v1.ObjectKey volume_id = 3;
+
+    int64 max_io_qps = 4;
+
+    string serial_number = 5;
 }
 
 message VirtioBlkCreateRequest {
-    VirtioBlk controller = 1;
+    VirtioBlkController controller = 1;
 }
 
 message VirtioBlkCreateResponse {
@@ -37,7 +43,7 @@ message VirtioBlkCreateResponse {
 }
 
 message VirtioBlkDeleteRequest {
-    int64 controller_id = 1;
+    common.v1.ObjectKey id = 1;
 }
 
 message VirtioBlkDeleteResponse {
@@ -45,7 +51,7 @@ message VirtioBlkDeleteResponse {
 }
 
 message VirtioBlkUpdateRequest {
-    VirtioBlk controller = 1;
+    VirtioBlkController controller = 1;
 }
 
 message VirtioBlkUpdateResponse {
@@ -53,26 +59,26 @@ message VirtioBlkUpdateResponse {
 }
 
 message VirtioBlkListRequest {
-    int64 subsystem_id = 1;
+    // Intentionally empty.
 }
 
 message VirtioBlkListResponse {
-    repeated VirtioBlk controller = 1;
+    repeated VirtioBlkController controller = 1;
 }
 
 message VirtioBlkGetRequest {
-    int64 controller_id = 1;
+    common.v1.ObjectKey id = 1;
 }
 
 message VirtioBlkGetResponse {
-    VirtioBlk controller = 1;
+    VirtioBlkController controller = 1;
 }
 
 message VirtioBlkStatsRequest {
-    int64 controller_id = 1;
+    common.v1.ObjectKey id = 1;
 }
 
 message VirtioBlkStatsResponse {
-    int64 id = 1;
+    common.v1.ObjectKey id = 1;
     string stats = 2;
 }
--- a/storage/v1/frontend_virtio_scsi.proto
+++ b/storage/v1/frontend_virtio_scsi.proto
@@ -31,7 +31,7 @@ service VirtioScsiLunService {
 message VirtioScsiController {
     int64 id = 1;
     string name = 2;
-    NvmeControllerPciId pcie_id = 4;
+    PciEndpoint pcie_id = 3;
 }
 
 message VirtioScsiLun {
--- a/storage/v1/frontend_virtio_scsi.proto
+++ b/storage/v1/frontend_virtio_scsi.proto
@@ -5,21 +5,27 @@ syntax = "proto3";
 package opi_api.storage.v1;
 option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
 import "common.proto";
+import "object_key.proto";
 
 // Front End (host-facing) APIs.
 
 // Virtio-scsi emulation
 
-service VirtioScsiControllerService {
+service FrontendVirtioScsi {
+    rpc VirtioScsiTargetCreate (VirtioScsiTargetCreateRequest) returns (VirtioScsiTargetCreateResponse) {}
+    rpc VirtioScsiTargetDelete (VirtioScsiTargetDeleteRequest) returns (VirtioScsiTargetDeleteResponse) {}
+    rpc VirtioScsiTargetUpdate (VirtioScsiTargetUpdateRequest) returns (VirtioScsiTargetUpdateResponse) {}
+    rpc VirtioScsiTargetList   (VirtioScsiTargetListRequest)   returns (VirtioScsiTargetListResponse)   {}
+    rpc VirtioScsiTargetGet    (VirtioScsiTargetGetRequest)    returns (VirtioScsiTargetGetResponse)    {}
+    rpc VirtioScsiTargetStats  (VirtioScsiTargetStatsRequest)  returns (VirtioScsiTargetStatsResponse)  {}
+
     rpc VirtioScsiControllerCreate (VirtioScsiControllerCreateRequest) returns (VirtioScsiControllerCreateResponse) {}
     rpc VirtioScsiControllerDelete (VirtioScsiControllerDeleteRequest) returns (VirtioScsiControllerDeleteResponse) {}
     rpc VirtioScsiControllerUpdate (VirtioScsiControllerUpdateRequest) returns (VirtioScsiControllerUpdateResponse) {}
     rpc VirtioScsiControllerList   (VirtioScsiControllerListRequest)   returns (VirtioScsiControllerListResponse)   {}
     rpc VirtioScsiControllerGet    (VirtioScsiControllerGetRequest)    returns (VirtioScsiControllerGetResponse)    {}
     rpc VirtioScsiControllerStats  (VirtioScsiControllerStatsRequest)  returns (VirtioScsiControllerStatsResponse)  {}
-}
 
-service VirtioScsiLunService {
     rpc VirtioScsiLunCreate (VirtioScsiLunCreateRequest) returns (VirtioScsiLunCreateResponse) {}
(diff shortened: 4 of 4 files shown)


mkalderon:
Hi @benlwalker 
Few general questions: 
1. trying to understand nvme-of frontend. We have a network-device that once reaches the DPU, the storage commands are lost, we just have tcp / rdma packets, how will the DPU be able to intercept and provide middle / end services ? Or do you see the DPU in this case terminating the connection and establishing another ? more details on the use case and flow would be great. thanks.
3. again, would be great for a more specific example - do you see other use cases that will define an endpoint other than controller bdf which are not nvme-of ? 
thanks

glimchb:
> Hi @benlwalker Few general questions:
> 
> 1. trying to understand nvme-of frontend. We have a network-device that once reaches the DPU, the storage commands are lost, we just have tcp / rdma packets, how will the DPU be able to intercept and provide middle / end services ? Or do you see the DPU in this case terminating the connection and establishing another ? more details on the use case and flow would be great. thanks.
> 2. again, would be great for a more specific example - do you see other use cases that will define an endpoint other than controller bdf which are not nvme-of ?
>    thanks

I think the intent for TCP is to terminate NVMe/TCP connection from the Host and start and new one to the remote storage...
Whether it is useful for production or not - not for me to say, but it demonstrates the concept of having something in addition to PCIe well. not all vendors have to implement it... I would guess Virtio-blk will be added to this Device APIs...

I will let @benlwalker correct me if my understanding is wrong...

benlwalker:
We've so far only discussed the use of an xPU to offload the initiator side. The TCP endpoint I have in this proposal is instead to offload the target side, where the front-end NVMe host is over the network and the backend storage is attached to the xPU or to the host server the xPU is attached to. I included it primarily to get everyone thinking about a broader set of use cases.

mkalderon:
> We've so far only discussed the use of an xPU to offload the initiator side. The TCP endpoint I have in this proposal is instead to offload the target side, where the front-end NVMe host is over the network and the backend storage is attached to the xPU or to the host server the xPU is attached to. I included it primarily to get everyone thinking about a broader set of use cases.

thanks for the clarification

glimchb (changes requested):
Following the online discussion today, the PR will change to keep the original naming specific for NVMe

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -44,25 +40,14 @@ message NVMeSubsystem {
     // object's unique identifier
     common.v1.ObjectKey id = 1;
 
-    // NVMe subsystem NQN to which the controller belongs
-    // Refer to the NQN format in the NVMe base specifications, must not
-    // exceed 'NSV_NVME_SUBSYSTEM_NQN_LEN' bytes
-    string nqn = 2;
```
benlwalker, line 50: local PCI NVMe devices do not need a subsystem NQN

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
-    NvmeControllerPciId pcie_id = 4;
+    PciEndpoint pcie_id = 3;
 
     // maximum host IO queue pairs allowed, value will default to
     // limits in PCI device configuration; if set to 0 or more it
     // will default to maximum permitted per xPU's capability
-    uint32 max_io_qps = 5;
+    uint32 max_io_qps = 4;
 
     // maximum Number of namespaces that will be provisioned under
```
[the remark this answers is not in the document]
glimchb: +1 for adding clarification @jainvipin @mkalderon @orendu WDYT ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
     // must not be reused under the same subsystem
     uint32 nvme_controller_id  = 2;
 
-    // subsystem information
```
glimchb, line 76: Don't you need to specify which Subsystem this CTRL belongs if you many multiple Subsystems...?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
+    uint32 max_ns = 5;
 }
 
 message NVMeNamespace {
     // namespace's unique key
-    // replaces: int64 id = 1;
     common.v1.ObjectKey id = 1;
 
-    // subsystem for this namespace
-    common.v1.ObjectKey subsystem_id = 2;
```
glimchb, line 98: Don't you need to specify which Subsystem this NS belongs if you many multiple Subsystems...?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -261,8 +220,7 @@ message NVMeNamespaceUpdateResponse {
 }
 
 message NVMeNamespaceListRequest {
-    common.v1.ObjectKey subsystem_id = 1;
```
glimchb, line 264: can you list all NSs in the specific subsystem ?

storage/v1/frontend_virtio_blk.proto
```diff
@@ -19,60 +19,65 @@ service VirtioBlkService {
...
+    // The PCI endpoint where this device should appear
+    PciEndpoint pcie_id = 2;
+
+    // The back/middle-end volume to back this controller.
+    common.v1.ObjectKey volume_id = 3;
+
+    int64 max_io_qps = 4;
+
+    string serial_number = 5;
 }
```
glimchb, line 34: is there a `model_number` in virtio ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
-    common.v1.ObjectKey subsystem_id = 2;
-
-    // key of the PCIe controller object that will host this namespace.
-    common.v1.ObjectKey controller_id = 3;
-
-    // NSID present to the host by the NVMe PCIe controller.
-    // If not provided, then the controller will assign an unused NSID
-    // within the max namespace range - auto assigned nsid may not work
-    // for live migration
-    uint32 host_nsid = 4;
```
glimchb, line 107: I think, we need to be able to set NSID, same as you did below with LUN_ID...

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
-
-    // Block size in bytes, must be power of 2 and must be less than the max
-    // io size supported. Typically tested values are 512, and 4k.
-    int64 block_size = 5;
```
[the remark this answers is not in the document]
glimchb: Not sure I agree... 
for example see https://www.mankier.com/1/nvme-format#--block-size
we able to change block size programmatically...
not sure we need middle layer just to do this...
will let others to chime in @mkalderon @jainvipin @orendu

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
-
-    // Size/Capacity of the namespace in blocks, size in bytes will
-    // be BlockSize x NumBlocks.
-    int64 num_blocks = 6;
```
glimchb, line 115: same as above... plus I'm thinking AIO bdev from file to block... where we specify size? do we need split-like middle end just for this ?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
     // 64bit Extended unique identifier for the namespace
     // mandatory if guid is not specified, optional otherwise
-    fixed64 eui64 = 8;
+    fixed64 eui64 = 3;
 
     // Globally unique identifier for the namespace
-    common.v1.Uuid uuid = 9;
-
-    // reference to encryption key for the data at rest encryption
-    common.v1.ObjectKey crypto_key_id = 10;
```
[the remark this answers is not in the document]
glimchb: +1 `volume_id` below seems like a more generic name

storage/v1/frontend_nvme_pcie.proto
glimchb: I see now... 
you moved it to `NVMeNamespaceCreateRequest` like `uint32 nsid = 3;`

storage/v1/frontend_nvme_pcie.proto
glimchb: I see now... you moved it to `NVMeControllerCreateRequest`

storage/v1/frontend_nvme_pcie.proto
glimchb: I see now... you moved it to `NVMeNamespaceCreateRequest`

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -238,6 +191,12 @@ message NVMeControllerStatsResponse {
 
 message NVMeNamespaceCreateRequest {
     NVMeNamespace namespace = 1;
+
+    common.v1.ObjectKey subsystem_id = 2;
```
glimchb, line 195:
```suggestion
    // The subsystem to associate this controller with
    common.v1.ObjectKey subsystem_id = 2;
```

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: What you linked sends an NVMe format command. That's sent by the host system/user of the device to change the block size, which is very different than this administrative step. A format command would flow through the front-end all the way to the back end and change the block size there.

Emulating block sizes is not at all simple. For example:

- Emulating 4K blocks on top of 512 physical blocks has to deal with NVMe's atomicity guarantees, which cannot be less than 1 block.
- Emulating 512 byte blocks on top of 4K physical requires read/modify/write.

This is very much not just setting a number. It's a ton of logic to do this correctly. It cannot appear as a parameter here..

storage/v1/frontend_nvme_pcie.proto
benlwalker: I wasn't entirely sure how to do these "list" commands for namespaces. You can do it several different ways:

1) List all the namespaces in a subsystem
2) List all the namespaces in a subsystem visible to a particular host (i.e. through a specific controller)
3) List all namespaces globally

I don't have a preference.

storage/v1/frontend_virtio_blk.proto
benlwalker: I don't believe so. Now that I look I don't think it even has a serial number. This is what the 1.2 spec says is reported to the host:

```
struct virtio_blk_config { 
        le64 capacity; 
        le32 size_max; 
        le32 seg_max; 
        struct virtio_blk_geometry { 
                le16 cylinders; 
                u8 heads; 
                u8 sectors; 
        } geometry; 
        le32 blk_size; 
        struct virtio_blk_topology { 
                // # of logical blocks per physical block (log2) 
                u8 physical_block_exp; 
                // offset of first aligned logical block 
                u8 alignment_offset; 
                // suggested minimum I/O size in blocks 
                le16 min_io_size; 
                // optimal (suggested maximum) I/O size in blocks 
                le32 opt_io_size; 
        } topology; 
        u8 writeback; 
        u8 unused0; 
        u16 num_queues; 
        le32 max_discard_sectors; 
        le32 max_discard_seg; 
        le32 discard_sector_alignment; 
        le32 max_write_zeroes_sectors; 
        le32 max_write_zeroes_seg; 
        u8 write_zeroes_may_unmap; 
        u8 unused1[3]; 
        le32 max_secure_erase_sectors; 
        le32 max_secure_erase_seg; 
        le32 secure_erase_sector_alignment; 
};
```

storage/v1/frontend_nvme_pcie.proto
glimchb: we can do all 3...
- if client provided subsystem parameter -> List all the namespaces in a subsystem
- if client provided subsystem and controller parameters -> List all the namespaces in a subsystem only visible to specific controller
- if no parameters -> List all the namespaces globally

WDYT ?

storage/v1/frontend_nvme_pcie.proto
benlwalker: Well you'd create a back-end volume, and if you were creating an AIO volume backed by a file you could specify the size there.

glimchb:
@llabordehpe for some reason can't add you to reviewers list, so tagging you here. please review

storage/v1/frontend_virtio_blk.proto
```diff
@@ -10,7 +10,7 @@ import "common.proto";
```
glimchb, line 10: missing
```
import "object_key.proto";
```

storage/v1/frontend_virtio_scsi.proto
```diff
@@ -10,16 +10,21 @@ import "common.proto";
```
glimchb, line 10: missing
```
import "object_key.proto";
```

storage/v1/frontend_virtio_scsi.proto
```diff
@@ -28,28 +33,91 @@ service VirtioScsiLunService {
...
+message VirtioScsiTargetCreateResponse {
+    // Intentionally empty.
+}
+
+message VirtioScsiTargetDeleteRequest {
+    common.v1.ObjectKey target_id = 1;
+}
+
+message VirtioScsiTargetDeleteResponse {
+    uint32 result = 1;
```
llabordehpe, line 73: Why do we need a result field?

storage/v1/frontend_virtio_scsi.proto
```diff
@@ -28,28 +33,91 @@ service VirtioScsiLunService {
...
+}
+
+message VirtioScsiTargetUpdateRequest {
+    VirtioScsiTarget target = 1;
+}
+
+message VirtioScsiTargetUpdateResponse {
+    uint32 result = 1;
```
llabordehpe, line 81: Why do we need a result field?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -44,25 +40,14 @@ message NVMeSubsystem {
...
 
     // maximum namespaces within a subsystem
-    int64 max_ns = 5;
-
-    // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
-    string firmware_revision  = 6;
-
-    // FRU identfier, 16bytes opaque identity for the type of unit
-    bytes fru_guid = 7;
+    int64 max_ns = 4;
```
llabordehpe, line 50: Is there a reason why max_ns is type int64 in NVMeSubsystem, and type uint32 in NVMeController ?

storage/v1/frontend_virtio_scsi.proto
benlwalker: Just matches the NVMe equivalent now. I don't have an opinion.

storage/v1/frontend_nvme_pcie.proto
benlwalker: Copied from NVMe. But even there I think it should only be 16 bit.

glimchb: there is no 16 bit in grpc... :-(

storage/v1/frontend_virtio_scsi.proto
glimchb: in a separate PR we will replace all return values according to https://google.aip.dev/133

glimchb: opened #174 to track it after this PR merged

storage/v1/frontend_nvme_pcie.proto
mkalderon: - perhaps a unify/non-unify question, but if we use the same message for nvme-oF we will need the nqn, not worth having just one struct ? 
- NQN is still provided in the response of get ctrl info, perhaps we want to enable control for what this response will be

jainvipin:
> 

Would the frontend APIs look same for what's needed on the initiator vs. target?

storage/v1/frontend_nvme_pcie.proto
jainvipin: should it be uint32 then, instead of int64?

storage/v1/frontend_nvme_pcie.proto
jainvipin: Well if we want to keep the REST semantics - there are resources and then there are CRUD operations on them (using POST/PUT/GET/etc.). This allows all the information about resource to be fully kept in the resource itself. Moving subsystem reference out of the resource and putting in context of the create request will not keep the APIs as RESTful. The disadvantages could be:
- user intent is captured outside the resource definition i.e. a GET on an object may not reveal what was created. It is not a good design, usually someone should be able to do GET on all objects to reconstruct the entire configured intent.
- you need yet another mechanism to describe this relationship e.g. which subsystem a namespace corresponds to.
- Mixing actions with resource can set precedence of doing this for other fields too. Ref: https://en.wikipedia.org/wiki/Restful_Objects
- Breaking the assumptions can also make plenty of tooling around REST APIs unusable.

I'd suggest re-considering this decision and keeping NVMeSubsystem, NVMeController and NVMeNamespaces as resources that user can configure, including any relationships between them is specified through those object definitions. Please see any popular open source project (e.g. K8s object definitions) for reference.

storage/v1/frontend_nvme_pcie.proto
jainvipin: same comment as for subsystem_id, let's keep this field here.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
jainvipin: for some reason. I don't see that change, can we keep this field as
```
common.v1.ObjectKey services_keychain_id

File: storage/v1/frontend_nvme_pcie.proto [hunk 17]
```diff
@@ -189,6 +139,9 @@ message NVMeSubsystemStatsResponse {
 
 message NVMeControllerCreateRequest {
     NVMeController controller = 1;
+
+    // The subsystem to associate this controller with
+    common.v1.ObjectKey subsystem_id = 2;
```
jainvipin, line 144: suggest removing contextual information in REST actions.

storage/v1/frontend_nvme_pcie.proto
jainvipin, line 195: suggest removing contextual information in REST actions.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -238,6 +191,12 @@ message NVMeControllerStatsResponse {
 
 message NVMeNamespaceCreateRequest {
     NVMeNamespace namespace = 1;
+
+    common.v1.ObjectKey subsystem_id = 2;
+
+    // The namespace ID inside of the subsystem where this namespace
+    // can be found.
+    uint32 nsid = 3;
```
jainvipin, line 199: suggest removing contextual information in REST actions.

storage/v1/frontend_nvme_pcie.proto
jainvipin: Although I don't have a strong opinion on this. Usually List() APIs don't provide filter parameters, but if we do the semantics of filtering can be tricky e.g. find me with `A || B` or `A && B` or `(A && B) || C`. I know we want to keep things simple, but that's the most generic way of doing List filtering. I am okay if we enumerate common field(s) but still the interpretation of those fileds (togher) will be some implicit  semantic (i.e.. AND or OR operation) - which is not a good thing.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -13,25 +13,21 @@ import "object_key.proto";
 
 // NVMe/PCIe emulation
 
-service NVMeSubsystemService {
+service FrontendNvme {
```
jainvipin, line 16: +1 on combining these multiple services into one on various objects - it is just too much to have to deal with them separatelly.

jainvipin (changes requested):
@benlwalker - good to see various services combined for Nvme, Virtio etc. together. Some comments on the structural changes here; will provide nvme specific comments subsequently.

storage/v1/frontend_nvme_pcie.proto
benlwalker: The reason I changed it was because the previous code did not correctly capture all of the complexities of the possible relationships. My new approach does. But I understand your points here as well. Here's some of the possible issues with how it was done before:

1) If we ever want to support dispersed namespaces, then a namespace can be in more than one subsystem. That means the it can't hold just one subsystem_id.
2) Namespaces can be visible through more than one controller, so the namespace can't have just one controller_id.
3) A namespace can exist at different NSIDs for different controllers (and in different subsystems if we support dispersed namespaces)

How would you propose we solve these issues? Making the user fill out the entire list of relationships any time they want to make a call is not convenient at all.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: I think @glimchb and I are very much saying no, we cannot keep it as anything related to crypto. Is there a reason you absolutely must get the crypto information right here, and not get it by looking at the associated volume backing this namespace?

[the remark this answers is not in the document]
glimchb: I don't think @jainvipin disagrees, he is just not seeing the change on github from all the comments...
@benlwalker changed
```
    // reference to encryption key for the data at rest encryption
    common.v1.ObjectKey crypto_key_id = 10;
```
to
```
    // The back/middle-end volume to back this namespace.
    common.v1.ObjectKey volume_id = 5;
```

storage/v1/frontend_nvme_pcie.proto
jainvipin, line 50: if we remove nqn here how would host see a subsystem. @chaitanyahuilgol - any comments?

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -44,25 +40,14 @@ message NVMeSubsystem {
...
 
     // model number, must not exceed 'NSV_CTRLR_MODEL_NO_LEN' bytes
-    string model_number = 4;
+    string model_number = 3;
 
     // maximum namespaces within a subsystem
-    int64 max_ns = 5;
-
-    // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
-    string firmware_revision  = 6;
```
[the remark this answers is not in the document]
jainvipin: ack - we could remove this from here; however it is a good candidate for why we need to resolve #180 to remove this from spec and put in status.

storage/v1/frontend_nvme_pcie.proto
jainvipin: `subsystem_id` that a controller refers to should be in the `NVMeController` object for keeping resource definitiosn wholesome (i.e. an Nvmecontroller is not defined properly if it doesn't refer to a subsystem)

storage/v1/frontend_nvme_pcie.proto
jainvipin: ok I understand your point of view; can we deal with it using one of the following ways:
1. Create another namespace, attach to a different subsystem_id, but have the same `guid` for both. This could be done in the proposed APIs without any changes.
2. Alternatively. one can make subsystem_id a list within a `Namespace` object and so user can configure various subsystem it would need to correspond.

For example, k8s captures 1:n relationship typically by having a list of named references or selection (m:n) references.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
jainvipin: We use it for what you mentioned here - it helps allocate appropriate resources. So I'd not say `less flexible` but `optimized resource allocater` :-)

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
 message NVMeNamespace {
     // namespace's unique key
-    // replaces: int64 id = 1;
     common.v1.ObjectKey id = 1;
 
-    // subsystem for this namespace
-    common.v1.ObjectKey subsystem_id = 2;
-
-    // key of the PCIe controller object that will host this namespace.
-    common.v1.ObjectKey controller_id = 3;
```
jainvipin, line 101: let's not remove named references out of object resources.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
jainvipin: well if `block_size` is presented to the host for a block device, then shouldn't frontend should have this value? @chaitanyahuilgol any comments?

storage/v1/frontend_nvme_pcie.proto
jainvipin: well if you are virtualizing a block device, this is really a frontend parameter that user should configure on what to expose.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
jainvipin: I am good with keeping it `volume_id` and have crypto and middleend fields in that. Then `volume_id` really is a refrerence to middle_end object handle that contains various parameters for crypto/etc.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -73,70 +58,35 @@ message NVMeController {
...
-    common.v1.Uuid uuid = 9;
-
-    // reference to encryption key for the data at rest encryption
-    common.v1.ObjectKey crypto_key_id = 10;
-
-    // optimal write size hint to host driver. Host IO stack may use
-    // this to regulate IO size. Must be a multiple of the preferred write
-    // granularity. Must not exceed the controller maximum IO size value
-    // configured in the nvme agent config file.
-    uint32 optimal_write_size = 11;
```
jainvipin, line 134: Per spec, xPU exposes this value to the driver for optimal writes, so this would be a configuration to XPU. @chaitanyahuilgol - any comments? Btw, @benlwalker - are you suggesting these fields move to a backend object that a frontend object refers to but ultimately provided to the driver using the value in the referred backend object?

jainvipin (changes requested):
@benlwalker - thanks for the PR; some comments inline if we can agree to those!

jainvipin:
> We've so far only discussed the use of an xPU to offload the initiator side. The TCP endpoint I have in this proposal is instead to offload the target side, where the front-end NVMe host is over the network and the backend storage is attached to the xPU or to the host server the xPU is attached to. I included it primarily to get everyone thinking about a broader set of use cases.

@benlwalker - I am curious if the changes you made were exclusively for Target side (or Initiator side as well)? Of course the endpoint object is not really there.

storage/v1/frontend_nvme_pcie.proto
chaitanyahuilgol: The nqn provided here is intended to be reported in the identify controller response - SUBNQN field. @benlwalker can you please let us know what are the alternatives for getting the SUBNQN if not provided here?

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
mkalderon: We see this as NN value. true that current mainstream drivers don't use the NN value. But there are some bios drivers which look at this value. There could be a use-case where host boots from nvme-device using such bios drivers.

storage/v1/common.proto
```diff
@@ -6,22 +6,13 @@ syntax = "proto3";
...
 option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
 
-// The controller PCI-ID is used to address a given virtual controller. Virtual
-// controllers are organized into devices with Physical functions and SRIOV
-// virtual function within the physical functions. Currently, xPUs may
-// expose multiple devices with one physical function each and one or more
-// virtual functions under the physical function.
-message NvmeControllerPciId {
-    // Bus number, provided for future usage if needed. Currently set to ’0’
-    uint32 bus              = 1;
```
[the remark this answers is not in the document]
benlwalker: How does the xPU pick the bus number? I'm fairly certain it's picked by the OS PCI driver.

storage/v1/common.proto
```diff
@@ -6,22 +6,13 @@ syntax = "proto3";
...
-
-    // Device number, based on the NVMe device layout
-    uint32 device           = 2;
```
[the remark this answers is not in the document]
benlwalker: We do not want to preclude any topology. Multiple PFs with multiple VFs must be supported. My change isn't modifying that requirement, it's just acknowledging that the xPU can't control what BDF it's devices show up at entirely - they're picked by the OS PCI driver.

storage/v1/common.proto
```diff
@@ -6,22 +6,13 @@ syntax = "proto3";
...
-    // SRIOV Virtual function within the Device and Physical function.
-    // Set to 0 for Physical Function. Virtual Function numbering starts from 1
-    uint32 virtual_function  = 4;
+// Many front-ends expose PCI devices to the host. This represents that endpoint.
+// The Bus:Device:Function numbers will be assigned by the OS PCI driver, so we
+// cannot set those directly here. What we can pick is which physical function
+// and virtual function the xPU uses.
+message PciEndpoint {
+    // Physical function index
+    uint32 physical_id = 1;
```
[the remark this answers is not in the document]
benlwalker: Sure

storage/v1/frontend_nvme_pcie.proto
benlwalker: There is no consumer of the SUBNQN value for a local NVMe device as far as I'm aware. The xPU can generate a value to fill that field in with. Did you have a particular use case where this is used and must be configured by the administrator?

(My comments do not apply to NVMe-oF where this is definitely needed)

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: I think I understand what you're getting at. I was treating this object as an INPUT from the client. But this object is pulling double-duty as both an input and an output. I think we need to resolve the lack of clarity on that as part of #180. I'll comment there.

storage/v1/frontend_nvme_pcie.proto
benlwalker: In this case yes, the controller always refers to exactly one subsystem so we could leave this id if we really wanted to maintain the relationship encoding in the objects.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: You should absolutely not use NN for resource provisioning. You shouldn't use it at all. I suspect it won't be entirely uncommon for NN to be set to the maximum value, to allow an unknown number of namespaces to be added or removed from a subsystem in the future. You cannot change NN without a reset, so most implementations should err on the side of going very large here.

Both the Linux kernel drivers and the SPDK drivers have been burned by this and have had very deep discussions around NN. The result of those discussions is do not use NN for anything, especially provisioning resources. We cannot re-introduce this mistake to OPI.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: I would see it as a major design flaw if we have parameters on the front-end that the user has to fill out that only exist to be validated against an automatic calculation based on the back-end volume. That's just adding unnecessary complexity for the user of this API.

This number cannot be an input from the user during the front-end configuration.

storage/v1/frontend_nvme_pcie.proto
benlwalker: We should allow that flexibility, either creating a middle-end API to truncate an arbitrary volume, or by allowing it to be configured directly in the back-end. But it cannot go here.

storage/v1/frontend_nvme_pcie.proto
benlwalker: I am suggesting that this configuration be done when creating the back-end/middle-end objects, which are what actually dictate these values. The front-end just reports them via the NVMe interfaces.

storage/v1/common.proto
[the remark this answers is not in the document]
jainvipin: I am agreeing that this is not set by xPU. that's why it is reported (in `Status`) by what is seen by xPU, not configured/set by xPU.

storage/v1/frontend_nvme_pcie.proto
chaitanyahuilgol: @benlwalker we can make this optional. but we should keep it around, we should give the client flexibility to configure this if they choose to

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
chaitanyahuilgol: IMO, this also qualifies for an optional parameter, we should not enforce any particular implementation by not allowing  this configuration. If clients want to provide and/or xPU want to use this, then it should be available.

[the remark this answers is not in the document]
mkalderon: > IMO, this also qualifies for an optional parameter, we should not enforce any particular implementation by not allowing this configuration. If clients want to provide and/or xPU want to use this, then it should be available.

agree

storage/v1/frontend_nvme_pcie.proto
benlwalker: As we're headed now, we likely will not unify front-end NVMe-oF with this.

benlwalker: Do you or does anyone have any use case for this? Again, the drivers do not even look at it. I'm trying to understand the motivation for wanting it. If there's some reason to have it I'm not at all opposed, but let's please provide reasons for things to exist in this API or we'll end up adding everything in the entire specification.

storage/v1/frontend_nvme_pcie.proto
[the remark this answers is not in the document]
benlwalker: NN is the maximum namespace id allowed, not the maximum number of namespaces attached. For example, you can support 4 namespaces with a maximum namespace ID value of 1000. Those namespaces can be at nsid 0, 12, 569, and 985.

We need these things to be documented very clearly. If this is NN, no implementation should use this value to do any internal resource provisioning. Lots of users are going to set this to MAXUINT16 or whatever the max is, and rightly so. There are some BIOS and other "broken" (old, really) NVMe drivers that use NN for namespace enumeration, so it does of course make sense to expose devices with NN values that are far smaller as well.

I believe we should really have two configuration values. The maximum NSID (NN) and the maximum number of namespaces. Both can be optional.

storage/v1/frontend_nvme_pcie.proto
benlwalker: I suspect this is going to simplify after #187 because the "list" of associated controllers (and maybe subsystems) will actually be in the status, not the spec.

PR #175, opened 2022-10-28
Split CQ/SQ and add max queue entries

amarv-marvell:
1) As per spec multiple submission queues may utilize the same completion queue.
   Hence decoupling the two.
2) Submission queues and completion queues can have different depths.
Will push poc updates once poc changes for Vipin's request are uploaded.

Signed-off-by: Amar Vishwakarma <amarv@marvell.com>


glimchb:
@benlwalker @llabordehpe @jkogutINTC @intelfisz @GottliebNoam please review

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,10 +79,21 @@ message NVMeController {
     // xPU's PCI ID for the controller
     NvmeControllerPciId pcie_id = 4;
 
-    // maximum host IO queue pairs allowed, value will default to
+    // maximum number of host submission queues allowed, value will default to
     // limits in PCI device configuration; if set to 0 or more it
     // will default to maximum permitted per xPU's capability
-    uint32 max_io_qps = 5;
+    uint32_t max_nsq = 5;
```
glimchb, line 85:
```
frontend_nvme_pcie.proto:85:5: "uint32_t" is not defined.
frontend_nvme_pcie.proto:90:5: "uint32_t" is not defined.
frontend_nvme_pcie.proto:93:5: "uint32_t" is not defined.
frontend_nvme_pcie.proto:96:5: "uint32_t" is not defined.
```
please fix

glimchb (changes requested):
failed CI, please fix `uint32_t`

glimchb (commented):
Thanks for the fixes, looks much better now.

Don't forget to run `make` and commit the generated files as well. this is why CI failing...

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,10 +79,21 @@ message NVMeController {
...
+
+    // maximum number of host completion queues allowed, value will default to
+    // limits in PCI device configuration; if set to 0 or more it
+    // will default to maximum permitted per xPU's capability
+    uint32_t max_ncq = 5;
+  
+    //maximum submission queue entries; given as index of 2
```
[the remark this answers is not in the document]
amarv-marvell: done

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: fixed

amarv-marvell:
> 

done

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,14 +79,27 @@ message NVMeController {
...
     // will default to maximum permitted per xPU's capability
-    uint32 max_io_qps = 5;
+    uint32 max_nsq = 5;
+
+    // maximum number of host completion queues allowed, value will default to
+    // limits in PCI device configuration; if set to 0 or more it
+    // will default to maximum permitted per xPU's capability
+    uint32 max_ncq = 6;
+  
+    // maximum submission queue depth; given as index of 2
```
benlwalker, line 92: Do you mean power of 2?

benlwalker (changes requested):

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: yes. corrected!

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,14 +79,27 @@ message NVMeController {
     // xPU's PCI ID for the controller
     NvmeControllerPciId pcie_id = 4;
 
-    // maximum host IO queue pairs allowed, value will default to
```
benlwalker, line 82: Only because these comments end up becoming the public documentation, I'm going to be very pedantic for a moment. In several places there is phrasing like this:

> value will default to limits in PCI device configuration; if set to 0 or more it will default to maximum permitted per xPU's capability

You didn't write that initially, but in this patch you do propagate that to several new values. That phrasing doesn't make any sense though, so let's fix it. I suggest:

> If not set, the xPU will provide a default.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,14 +79,27 @@ message NVMeController {
...
+    // maximum number of host submission queues allowed, value will default to
     // limits in PCI device configuration; if set to 0 or more it
     // will default to maximum permitted per xPU's capability
-    uint32 max_io_qps = 5;
+    uint32 max_nsq = 5;
+
+    // maximum number of host completion queues allowed, value will default to
```
benlwalker, line 87: Similarly here, I'd suggest:

> Maximum number of host completion queues allowed. Optional. If not set, the xPU will provide a default.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,14 +79,27 @@ message NVMeController {
...
+    // limits in PCI device configuration; if set to 0 or more it
+    // will default to maximum permitted per xPU's capability
+    uint32 max_ncq = 6;
+  
+    // maximum submission queue depth; given as power of 2
```
benlwalker, line 92: There's actually some subtlety to this. It should not say "queue depth", it really should say:

> Maximum number of submission queue entries per submission queue, as a power of 2.

This is not the same as queue depth because NVMe allows devices to consume off of the submission queue without reporting a completion immediately. So you can theoretically have a higher total queue depth of outstanding commands than entries in the submission queue. Very few devices actually work this way, but some really do.

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -79,14 +79,27 @@ message NVMeController {
...
+    // default value as per spec is 6
+    uint32 sqes = 7;
+
+    // maximum completion queue depth; given as power of 2
```
benlwalker, line 96: The completion queue side is simpler, but for consistency with the submission side I think we should write

> Maximum number of completion queue entries per completion queue, as a power of 2.

benlwalker (changes requested):

jainvipin (commented):
LGTM

glimchb:
@benlwalker good proposals on comments, agree that since this is going to be our public documentation, make sense to fix those to be as clear and as correct as possible, thanks for your time reviewing this and thanks for suggesting the alternative language. will be easy for @amarv-marvell to fix!

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: done.

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: done.

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: done.

storage/v1/frontend_nvme_pcie.proto
amarv-marvell: done.

amarv-marvell:
> 

done.

glimchb:
@benlwalker give it a look and if ok, I will merge it today

benlwalker (approved):

Issue #174, opened 2022-10-27
[storage] : Change APIs according to `Resource-oriented design` https://google.aip.dev/121


glimchb:
After #171 is done we need a separate PR to fix all storage according to https://google.aip.dev/121

- https://google.aip.dev/131 - Get
- https://google.aip.dev/132 - List
- https://google.aip.dev/133 - Create
- https://google.aip.dev/134 - Update
- https://google.aip.dev/135 - Delete

and https://google.aip.dev/127 - `HTTP and gRPC Transcoding`

llabordehpe:
Do you want to follow Google's AIPs just for Create (your link https://google.aip.dev/133), or also for Update, Delete, Get, etc... ?

glimchb:
yes, absolutely for everything @llabordehpe just gave `Create` as an example

glimchb:
After #218:
```
+--------------------------------------+------------------+----------------+
|                 RULE                 | TOTAL VIOLATIONS | VIOLATED FILES |
+--------------------------------------+------------------+----------------+
| core::0156::forbidden-methods        |               11 |              7 |
| core::0127::http-annotation          |               12 |              7 |
| core::0127::resource-name-extraction |               22 |              7 |
| core::0192::has-comments             |              380 |              8 |
+--------------------------------------+------------------+----------------+
```

PR #179, opened 2022-10-31
[storage][front-end] Delete xxxCreateResponse messages

glimchb:
Following Google [AIP-133](https://google.aip.dev/133) - Standard methods: Create

Related to #174


PWAlessi (approved):
Looks good, thanks for doing this!

mestery (approved):
STAMP

benlwalker (approved):

llabordehpe (approved):
Looks good.

Could this also be used to have the API provider assign the object handle,
if the API user passes in a 0-length handle in the XxxCreateRequest for example ?

glimchb:
> Looks good.
> 
> Could this also be used to have the API provider assign the object handle, if the API user passes in a 0-length handle in the XxxCreateRequest for example ?

@llabordehpe I think so yes. AIP is talking about this here https://google.aip.dev/133#user-specified-ids and here https://google.aip.dev/155

jainvipin:
LGTM - This change is inline with REST semantics where object itself is returned from a POST (aka create) request as mentioned [ here](https://google.aip.dev/133).

Having said the usefulness on passing the context back is desired upon failure. HTTP codes cover a full gammut of errors on the object resources as listed [here](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes), but in my experience, sometimes they are not sufficient or doesn't capture the domain specific context of an error. For now let's stick to the REST semantics and then discuss later if we must return any specific return message to handle failures better.

jainvipin (approved):

glimchb:
got almost all the responses, merging this now

PR #186, opened 2022-11-02
[storage][front-end] Add google.api.http

glimchb:
Following Google [AIP-127](https://google.aip.dev/127) - HTTP and gRPC Transcoding

and https://cloud.google.com/endpoints/docs/grpc/transcoding

Related to #174


PWAlessi (approved):
other than the url paths/names, looks good.

PWAlessi (changes requested):
url paths and names need to change

glimchb:
> url paths and names need to change

fixed

PWAlessi (approved):
I'm not positive that this is correct because I've never implemented protobuf over REST/http. I'd say that if the proto compiler works, it's probably fine to merge. I'd also add a backlog item for building out (at least a mock) http server to see if it works as expected.

glimchb:
> I'm not positive that this is correct because I've never implemented protobuf over REST/http. I'd say that if the proto compiler works, it's probably fine to merge. I'd also add a backlog item for building out (at least a mock) http server to see if it works as expected.

opened #188 to track backlog item

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -9,35 +9,102 @@ import "common.proto";
...
 service NVMeSubsystemService {
-    rpc NVMeSubsystemCreate (NVMeSubsystemCreateRequest) returns (NVMeSubsystem) {}
-    rpc NVMeSubsystemDelete (NVMeSubsystemDeleteRequest) returns (google.protobuf.Empty) {}
-    rpc NVMeSubsystemUpdate (NVMeSubsystemUpdateRequest) returns (NVMeSubsystem) {}
-    rpc NVMeSubsystemList   (NVMeSubsystemListRequest)   returns (NVMeSubsystemListResponse)   {}
-    rpc NVMeSubsystemGet    (NVMeSubsystemGetRequest)    returns (NVMeSubsystem)    {}
+    rpc NVMeSubsystemCreate (NVMeSubsystemCreateRequest) returns (NVMeSubsystem) {
+        option (google.api.http) = {
+            post: "/v1/subsystems"
+            body: "subsystem"
```
jainvipin, line 22: Do we want to prefix the path with `nvme`, `virtio`, etc. i.e. instead of implicit assumptions we can keep the nvme path to `/v1/nvme/subsystems`,`/v1/nvme/controllers`, etc.
Also there is an assumption that backend/frontend/middle API objects wouldn't conflict here if we don't include that in the path. i.e. `/v1/nvme/frontend/subsystems` or `/v1/nvme/fe/subsystems` vs. `/v1/nvme/subsystems`

storage/v1/frontend_virtio_blk.proto
```diff
@@ -6,17 +6,40 @@ package opi_api.storage.v1;
...
+        };
+    }
+    rpc VirtioBlkDelete (VirtioBlkDeleteRequest) returns (google.protobuf.Empty) {
+        option (google.api.http) = {
+            delete: "/v1/virtioblks/{virtioblk}"
+        };
+    }
+    rpc VirtioBlkUpdate (VirtioBlkUpdateRequest) returns (VirtioBlk) {
+        option (google.api.http) = {
+            patch: "/v1/virtioblks"
```
jainvipin, line 29: inline with previous comment this could be `/v1/virtio/blocks`, etc. or `/v1/virtio/frontend/blocks`?

jainvipin (commented):
one comment @glimchb on the path hierarchy

storage/v1/frontend_nvme_pcie.proto
glimchb: yes, that' good suggestion. Will fix it on follow-up PR! great

llabordehpe:
Shouldn't the objects "hierarchy" be part of the the URLs (like in this [example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/endpoints/bookstore-grpc/http_bookstore.proto))?
For example, in NVMeControllerCreate, don't we want:
   post: "/v1/subsystems/{subsystem}/controllers"
   body: "controller"
instead of :
   post: "/v1/controllers"
   body: "controller"

glimchb:
> Shouldn't the objects "hierarchy" be part of the the URLs (like in this [example](https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/endpoints/bookstore-grpc/http_bookstore.proto))? For example, in NVMeControllerCreate, don't we want: post: "/v1/subsystems/{subsystem}/controllers" body: "controller" instead of : post: "/v1/controllers" body: "controller"

Yes, it should. I just didn't want to overload this PR with hierarchy since it can lead to a heavy discussion on who is parent of who... I opened #188 to track the correct hierarchy fixes

mestery (approved):

storage/v1/frontend_nvme_pcie.proto
```diff
@@ -9,35 +9,102 @@ import "common.proto";
...
+        };
+    }
+    rpc NVMeSubsystemDelete (NVMeSubsystemDeleteRequest) returns (google.protobuf.Empty) {
+        option (google.api.http) = {
+            delete: "/v1/subsystems/{subsystem}"
+        };
+    }
+    rpc NVMeSubsystemUpdate (NVMeSubsystemUpdateRequest) returns (NVMeSubsystem) {
+        option (google.api.http) = {
+            patch: "/v1/subsystems"
```
llabordehpe, line 32: I think it shoud have been:
   patch: "/v1/subsystems/{subsystem}"

mkalderon:
@glimchb late response. looks good.

PR #190, opened 2022-11-07
[storage] Rename NvmeControllerPciId to PciEndpoint

benlwalker:
This is an important change. The OPI configuration APIs cannot pick the bus:device:function - that's assigned by the OS PCI driver. We can only pick the physical and virtual function numbers.


storage/v1/common.proto
```diff
@@ -6,22 +6,13 @@ syntax = "proto3";
...
 option go_package = "github.com/opiproject/opi-api/storage/v1/gen/go";
 
-// The controller PCI-ID is used to address a given virtual controller. Virtual
-// controllers are organized into devices with Physical functions and SRIOV
-// virtual function within the physical functions. Currently, xPUs may
-// expose multiple devices with one physical function each and one or more
-// virtual functions under the physical function.
-message NvmeControllerPciId {
-    // Bus number, provided for future usage if needed. Currently set to ’0’
-    uint32 bus              = 1;
```
jainvipin, line 16: we can add this to the status, perahps?

storage/v1/common.proto
```diff
@@ -6,22 +6,13 @@ syntax = "proto3";
...
-
-    // Device number, based on the NVMe device layout
-    uint32 device           = 2;
```
jainvipin, line 19: would prefer keeping this in, as some xPUs can offer multiple devices and user could allow selecting the device to offer this on. Note that there are multi-host deployment of xPUs and in some cases user can select exposing separate devices.

jainvipin (changes requested):
some comments inline @benlwalker

storage/v1/common.proto
PWAlessi, line 16: Is there a case where a user of the api needs the bus and device for some other functionality?

benlwalker: There's a lot of cases where it would be useful, but it doesn't matter because the xPU doesn't know these values. They're assigned by the OS.

benlwalker: How does the xPU know these values?

storage/v1/common.proto
benlwalker: This change does not preclude multiple physical or virtual functions in any way.

storage/v1/common.proto
jainvipin: yes not generically - this may be the case only if you have trusted host. so we can ignore this comment if this is not generic enough to report BDF assigned by the host in the status.

mkalderon (approved):

benlwalker:
Updated after latest discussion on Slack. I think it may be interesting to return these in the status too, just for "discovery" purposes. And we need some way to discover what the valid values for an xPU are. But we can do those in separate PRs.

storage/v1/common.proto
glimchb: we might need a new BDF struct on the backend APIs if we want to attach local NVMe devices to DPU and expose them to host

glimchb (changes requested):
committed `storage/googleapis/google/api/` by mistake @benlwalker ?

glimchb (approved):
see slack, agreement reached, merging this now

PR #194, opened 2022-11-08
[storage] Add java support

glimchb:


storage/v1alpha1/backend_aio.proto
```diff
@@ -3,6 +3,11 @@
 
 syntax = "proto3";
 package opi_api.storage.v1;
+
+option java_package = "opi_api.storage.v1";
```
seroyer, line 7: I think we should use consistent versioning between go and java.

glimchb: it is for now, the package is still `v1` see above the java line...
could not change it to alpha in the single previous PR
need new PR to complete the renaming

PR #204, opened 2022-11-14
Fixing namespace conflict

glimchb:
Fixing namespace conflict

`common` is too generic and we hit namespace conflict
when we try to use it.

see https://stackoverflow.com/questions/67693170/proto-file-is-already-registered-with-different-packages
see https://developers.google.com/protocol-buffers/docs/reference/go/faq#fix-namespace-conflict


mestery (approved):
image

Issue #188, opened 2022-11-03
backlog: re-check `google.api.http` URLs


glimchb:

jainvipin:
Recommend:
Nvme: `/v1alpha/nvme/frontend/...`
iSCSI: `/v1alpha/iscsi/frontend/...`
Virtio: `/v1alpha/virtio/frontend/...`
Middleend: `/v1alpha/middleend/...`

happy to submit a PR if we agree.

glimchb:
not frontend before the protocol ?

glimchb:
also see https://google.aip.dev/122#resource-uris
```
//calendar.googleapis.com/users/vhugo1802
vs
https://calendar.googleapis.com/v3/users/vhugo1802
```

PR #207, opened 2022-11-14
[storage] Add target to virtio-scsi, clean up virtio-blk

benlwalker:
virtio-scsi has the same 3 level object system as NVMe


glimchb (commented):
LGTM, thanks @benlwalker

glimchb:
@benlwalker you can add `Intel` `Copyright` to the files you changing...

PWAlessi (approved):
lgtm

storage/v1alpha1/frontend_virtio_scsi.proto
```diff
@@ -14,8 +14,37 @@ import "object_key.proto";
 import "google/protobuf/empty.proto";
 import "google/api/annotations.proto";
 
-// Front End (host-facing) APIs. Mostly used for Virtio-scsi emulation emulation and host presentation as alternative to Virtio-blk.
+// Front End (host-facing) APIs. Mostly used for Virtio-scsi emulation and host presentation as alternative to Virtio-blk.
 service FrontendVirtioScsiService {
+    rpc CreateVirtioScsiTarget (CreateVirtioScsiTargetRequest) returns (VirtioScsiTarget) {
+        option (google.api.http) = {
+            post: "/v1/virtioscsitargets"
```
jainvipin, line 21: URLs needs hierarchy - hopefully we can do it after this PR; the flattened url names in this and in NVMe protos can be done separately.

storage/v1alpha1/frontend_virtio_scsi.proto
```diff
@@ -75,15 +104,63 @@ service FrontendVirtioScsiService {
...
+    common.v1.ObjectKey target_id = 1;
+}
+
+message VirtioScsiTargetStatsRequest {
+    common.v1.ObjectKey target_id = 1;
+}
+
+message VirtioScsiTargetStatsResponse {
+    common.v1.ObjectKey id = 1;
+    string stats = 2;
```
jainvipin, line 163: not sure if this makes any sense, returning a big string for stats; if stats are not needed, we can remove it from here and introduce if with a bit more structure/thinking later.

jainvipin (commented):
one minor comment otherwise looks good

storage/v1alpha1/frontend_virtio_scsi.proto
glimchb: I opened #188 to track the fix of all the URLs at once

storage/v1alpha1/frontend_virtio_scsi.proto
glimchb: we have now `string stats` defined in all our objects, so I'm personally OK with this now to align with what we have already and then fix stats later on

glimchb (approved):
looks like straight forward alignment we doing already today for NVMe
opened separate issues to fix the review comments about URLs and STATs

storage/v1alpha1/frontend_virtio_scsi.proto
jainvipin: sure, that would work.

jainvipin (approved):
LGTM

PR #208, opened 2022-11-14
[storage] Add Crypto middle end volume

glimchb:


glimchb:
> Should the Crypto functionality be included in the middle end protobuf or should there be a separate common crypto capability that can be tied to the function (Storage, Security, Networking, etc.) that would create, delete, update, ... the capability?
> 
> Perhaps we can iterate on that aspect of the common functionality.

d@re (data at rest encryption) is only applicable for storage

we can definitely iterate on this common once we have another example that uses the same, we can easily move it around to a common place

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
+    rpc GetCrypto    (GetCryptoRequest)    returns (Crypto)    {
+        option (google.api.http) = {
+            get: "/v1/volumes/{volume}"
+        };
+    }
+    rpc CryptoStats  (CryptoStatsRequest)  returns (CryptoStatsResponse)  {}
+}
 
-// TODO: encryption, compression, raid, QoS, multipath, ...
+message Crypto {
```
benlwalker, line 49: Should we call it Crypto or CryptoVolume? Without the word "Volume" I think it leaves people room to misinterpret how this object really works. It's not the global crypto service, it's a specific volume (that may forward to another volume).

benlwalker (changes requested):

storage/v1alpha1/middleend.proto
glimchb: Sure, I can add the word `Volume` to it, I just have to do it across all RPCs as well

glimchb: @jainvipin @orendu @mkalderon @llabordehpe what do you think about naming of this before I jump to change it ?

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
+    CryptoType cipher = 4;
+
+}
+
+message CreateCryptoRequest {
+    Crypto volume = 1;
+}
+
+message DeleteCryptoRequest {
+    common.v1.ObjectKey volume_id = 1;
```
llabordehpe, line 68: Which id is this one from Crypto struct? id or volume_id ? I assume it is id. Maybe we should rename it crypto_id (and same in GetCryptoRequest, CryptoStatsRequest, etc.. to remove ambiguity.

glimchb: make sense. and WDYT about `Crypto` vs `CryptoVolume` proposal from @benlwalker on the struct name ?

storage/v1alpha1/middleend.proto
llabordehpe: How does this crypto object really work? After calling CreateCrypto(), can the original volume id still be used (in CreateNvmeNamespace for example), or should the new id be used ? And in term of calls order, what happens if you call CreateCrypto AFTER already using a volume with CreateNvmrNamespace ?

glimchb: yes, layered services like SPDK. bdev on top of bdev and you can use either one of them

glimchb: I will try to show the usage now, after I merge this and implement in our bridge example https://github.com/opiproject/opi-spdk-bridge/tree/main/server

glimchb:
opened #210 to think about better name, will try to implement now the code, to show the usage example

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
+import "object_key.proto";
+import "google/protobuf/empty.proto";
+import "google/api/annotations.proto";
 
-// Middle End (Storage Services) APIs.
+// Middle End (Storage Services) APIs. For example, encryption, compression, raid, QoS, multipath, ...
+service MiddleendService {
+    rpc CreateCrypto (CreateCryptoRequest) returns (Crypto) {
+        option (google.api.http) = {
+            post: "/v1/volumes"
```
jainvipin, line 21: the flat API structure is evident here `v1/volumes` is very generic, if we preceed it with `v1/middleend/volumes` that would decouple it from the backend volumes and overlap.

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
 
-// TODO: encryption, compression, raid, QoS, multipath, ...
+message Crypto {
+    common.v1.ObjectKey crypto_id = 1;
+
+    // The back/middle-end volume to back this volume
+    common.v1.ObjectKey volume_id = 2;
+
+    // Key to be used for encryption
+    bytes key = 3;
```
jainvipin, line 56: Is it a shared secret or a private key?

jainvipin: do we need to provide secret for `initial vector` or do we believe use of IV is perhaps a stale thing?

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
+    common.v1.ObjectKey crypto_id = 1;
+}
+
+message CryptoStatsRequest {
+    common.v1.ObjectKey crypto_id = 1;
+}
+
+message CryptoStatsResponse {
+    common.v1.ObjectKey crypto_id = 1;
+    string stats = 2;
```
jainvipin, line 95: same comment about stats as in other PRs

storage/v1alpha1/middleend.proto
```diff
@@ -9,7 +9,88 @@ option java_multiple_files = true;
...
+    // Key to be used for encryption
+    bytes key = 3;
+
+    // Cipher to use
+    CryptoType cipher = 4;
+
+}
+
+message CreateCryptoRequest {
+    Crypto volume = 1;
```
jainvipin, line 64: minor, why do we call the field `volume` it should be called `crypto`? e.g. what if we have other service applied on the volume, would that also be called `volume` in the create/update requests?

jainvipin (commented):
some comments @glimchb

storage/v1alpha1/middleend.proto
glimchb: agree, part of #188

storage/v1alpha1/middleend.proto
glimchb: no idea, this is just first draft, let's discuss

storage/v1alpha1/middleend.proto
intelfisz: @glimchb - encryption is a volume-level property. If we don't set it at volume creation timepoint (since this is middle-end API) then we need to make it clear in the documentation that CreateCrypto() should be called immediately after volume creation to avoid a data mess.

glimchb: good point @intelfisz 

maybe we need an option to create crypto/compress/... basically all middle-end services together with the actual volume itself... documentation is nice, but often not enough. APIs should indeed encourage doing this somehow... 

maybe https://google.aip.dev/233 ? WDYT ?

intelfisz: This is one of the reasons @benlwalker has been arguing for a higher abstraction level which we follow in SMA approach - either you split things into multiple calls (which may be used wrong way and require careful documentation/usage notes) or you just provide additional parameters to a single high-level call (like enabling crypto upon volume creation) which does it the proper way.

benlwalker: Yes exactly. I think an alternative to SMA style big single calls is some sort of transaction manager service where all calls can be associated with a transaction and then committed at once. But you need one or the other.

Issue #191, opened 2022-11-07
discuss `firmware_revision`


glimchb:
from slack:

Vipin Jain
 [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855692671799?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
can we also close on the firmware version?

Ben Walker
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855710354389?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
sure - can you clarify what information you're trying to convey through the NVMe firmware version?

Vipin Jain
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855852366459?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
the version of the control/datapath being run for NVMe, which is different from that of the device's firmware. In our case customer changes this information based on their own P4 and so they want us to specify what version is being run

Ben Walker
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855876491859?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
does this need to be discovered via NVMe commands by the host?

Vipin Jain
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855914066939?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
well nvme driver does report the firmware revision, so that's what needs to be shown there

Ben Walker
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855928487799?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
but the customers need to see it through the NVMe firmware version?

Ben Walker
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855951923559?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
they can't query the device config some other way?

Vipin Jain
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667855979649679?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
but firmware version is exactly for that reason, why use some other config or sideband way?

Ben Walker
  [1 hour ago](https://opi-project.slack.com/archives/C03Q50JUWAU/p1667856010622759?thread_ts=1667829743.692569&cid=C03Q50JUWAU)
[...]

PR #214, opened 2022-11-21
move firmware_version to spec from status

jainvipin:
addresses #191 pls review @glimchb @benlwalker 
I think it wouldn't be a bad idea to keep firmware_version in both spec and status, so for systems that don't want firwmare version to be dynamic, can allow learning about it using status. I can modify the PR to refelect if you all agree. For now I have kept it out ot status.


glimchb:
@jainvipin can you please explain the use case again to re-fresh the thread ?  
- who would configure this filed ? customer ? like cloud orchestrator ? 
- how one would see this filed afterwards ? like this ?
```
$ nvme list
Node             SN              Model                Namespace Usage                      Format           FW Rev
---------------- --------------- -------------------- --------- -------------------------- ---------------- --------
/dev/nvme4n1     MY_SERIAL         MY_MODEL             1         524.29  kB /   1.60  TB    512   B +  0 B   2.1.3
```

jainvipin:
closing this PR in favor of offline discussion; the better method to keep the firmware_revision is to tie it to the method delivering the firmware dynamically to the xPU and take it up in lifecycle provisioning work vs. storage APIs.

PR #216, opened 2022-11-28
Add Hostnqn to backend

glimchb:


storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -32,6 +32,7 @@ message NVMfRemoteController {
     NvmeMultipath multipath = 9;
     int64 io_queues_count = 10;
     int64 queue_size = 11;
+    string hostnqn = 12;
```
jainvipin, line 35: is this the target's host nqn?

jainvipin (commented):
one question @glimchb

storage/v1alpha1/backend_nvme_tcp.proto
glimchb: `string subnqn = 6;` is target NQN
`string hostnqn = 12;` is DPU NQN
it is part of the backend APIs because we want to be able to connect to different targets using different Host NQNs

benlwalker (approved):

seroyer (approved):

storage/v1alpha1/backend_nvme_tcp.proto
jainvipin: may be it is better to document the fields to avoid confusion!

PR #220, opened 2022-12-09
[storage] define statistics object

glimchb:


benlwalker (approved):

glimchb:
@ekaley FYI

storage/v1alpha1/opicommon.proto
```diff
@@ -37,6 +37,18 @@ message PciEndpoint {
...
+    int32    read_ops_count         = 2;
+    int32    write_bytes_count      = 3;
+    int32    write_ops_count        = 4;
+    int32    unmap_bytes_count      = 5;
+    int32    unmap_ops_count        = 6;
+    int32    read_latency_ticks     = 7;
+    int32    write_latency_ticks    = 8;
+    int32    unmap_latency_ticks    = 9;
+}
+
```
llabordehpe, line 51: Shoudn't these be 64-bit counters ?

llabordehpe:
Not sure I understand why there are VolumeStats in the NVMe subsystem, controller and namespace objects. Are the stats for the subsystem an aggregate for all volumes/namespaces currently (or ever) attached to that subsystem, versus the namespace stats which are for this volume only ?

glimchb:
> Not sure I understand why there are VolumeStats in the NVMe subsystem, controller and namespace objects. Are the stats for the subsystem an aggregate for all volumes/namespaces currently (or ever) attached to that subsystem, versus the namespace stats which are for this volume only ?

exactly, aggregate

jainvipin:
I like this approach, where stats are associated clearly with the object itself at a given point in time i.e. it is clear and works for not so heavy stats users. I am commenting in context of OTEL as the method of exporting/streaming the stats.
To @llabordehpe's point it would good to document the desired behavior on the Volume stats in various objects where the object is embeded.

glimchb:
> I like this approach, where stats are associated clearly with the object itself at a given point in time i.e. it is clear and works for not so heavy stats users. I am commenting vis-a-vis OTEL as the method of exporting/streaming the stats. 

Good topic to discuss, how exactly we want to stream telems from OTEL perspective. I would imagine we should compile DPU SDK with OTEL SDK and stream directly. So what we can do in OPI is to define the OTEL format and telems needed streaming...

> To @llabordehpe's point it would good to document the desired behavior on the Volume stats in various objects where the object is embeded.

Agree 100%, documentation is lacking.

storage/v1alpha1/opicommon.proto
glimchb: I guess, we can make them 64 bit

PR #223, opened 2022-12-12
Add Inventory protobuf in common area

sandersms:
Starting point for Inventory protobuf for use in reporting the Device information.


common/v1/inventory.proto
```diff
@@ -0,0 +1,36 @@
...
+    string serial_no = 9;
+    string part_no = 10;
+}
+
+message InventoryGetRequest {
+    // Empty
+}
+
+message InventoryGetResponse {
+    DeviceInfo devinfo = 1;
```
glimchb, line 35: I think we need to return 6 structs here:
- bios
- system
- baseboard
- chassis
- processor
- memory

glimchb (commented):
it's a good start. I think we need to follow DMI (smbios) tables and define those messages:
- bios
- system
- baseboard
- chassis
- processor
- memory
and then return those 6 in the `InventoryGetResponse`

glimchb (changes requested):
can you please rebase? I think it will fail on several linters now

common/v1/inventory.proto
```diff
@@ -0,0 +1,36 @@
...
+
+syntax = "proto3";
+
+package opi_api.inventory.v1;
+option go_package = "github.com/opiproject/opi-api/common/v1/gen/go";
+
+// Service functions for the device inventory data
+service InventorySvc {
+    // retrieves the inventory data for the device
+    rpc InventoryGet (InventoryGetRequest) returns (InventoryGetResponse) {}
```
glimchb, line 14: according to https://google.aip.dev/131 rpc name should be `GetInventory`
according to https://google.aip.dev/131 rpc response should be just `Inventory`
example:
```
    rpc GetInventory (GetInventoryRequest) returns (Inventory)    {
        option (google.api.http) = {
            get: "/v1/{name=inventory}/{inventory}"
        };
        option (google.api.method_signature) = "name";
    }
```

glimchb (approved):

common/v1/inventory.proto
sandersms: Rebased and cleaned up the lint reported checks that are active.

PR #224, opened 2022-12-13
document api extension methods

jainvipin:
@sandersms @glimchb  - pls review

I have tried out some very basic things; as we try to extend we can update the documentation if there are more things to consider.

### DIFF
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -13,11 +13,11 @@ Having limits on many features and scale limits may be generic, however a vendor
 a vendor or generically with a vendor. For example, exposing a limitation on the number of VFs a device should expose, is a limitation
 of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify as a generic limitation and doesn't
 belong in vendor specific extensions. On the other hand, if a vendor's product (say, an older generation of xPU) has a limitation on
-supporting VirtIO device emulation onf PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
+supporting VirtIO device emulation on PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
 - Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general
 consensus that this is a generic feature. For example, providing deeper IO telemetry (e.g. IO completion time, read/write failure
 codes) seems generic enough, but can start as a vendor extension (in Status part of an object), and can make into OPI API as it
-gains more adoption/support. On the other hand, supporting a custom programmable compression algorithm is likely goin to be
+gains more adoption/support. On the other hand, supporting a custom programmable compression algorithm is likely going to be
 vendor specific feature.
 - The boundaries on generic vs vendor specific extensions may look blur - there are some easy ones to fall on either side i.e.
 generic vs. vendor specific. The ones in the middle ground can start as vendor specific extensions making their way into
@@ -83,7 +83,7 @@ For improved backward compatibility of the OPI APIs
 - Vendor specific messages are enumerated from the values `1024` onwards
 - Specific fields are never included directly in the extensions, but a vendor specific messsage is include
 - Vendor specific message can have its own (and evolving) fields
-- The vendor speific messages are kept in the vendor specific directories, however in the above example, they are kept in the same
+- The vendor specific messages are kept in the vendor specific directories, however in the above example, they are kept in the same
 file for brevity of diffs.
 
 Note that [proto2 extensions](https://developers.google.com/protocol-buffers/docs/proto#extensions) are not
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -5,21 +5,33 @@ may vary, following guidelines can help determine if a new field/message should 
 native OPI field/message.
 
 - Vendor specific feature: If the proposed extension implements a vendor specific feature, then it should be added as an extension.
-A generic feature should be considered as part of the native OPI API in the respective areas (network, storage, storage, etc.).
+A generic feature should be considered as part of the native OPI API in the respective areas (network, storage, security, etc.).
 Note that all OPI API fields/messages may not be mandatory, allowing vendors to implement them as their product and/or software
 release process occurs
 - Vendor specific limitation: If the proposed extension exposes a limitation of a product, then it should be added as an extension.
 Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
-a vendor or generically with a vendor
-- Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general consensus
-that this is a generic feature
+a vendor or generically with a vendor. For example, exposing a limitation on the number of VFs a device should expose, is a limitation
+of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify as a generic limitation and doesn't
+belong in vendor specific extensions. On the other hand, if a vendor's product (say, an older generation of xPU) has a limitation on
+supporting VirtIO device emulation onf PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
+- Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general
+consensus that this is a generic feature. For example, providing deeper IO telemetry (e.g. IO completion time, read/write failure
+codes) seems generic enough, but can start as a vendor extension (in Status part of an object), and can make into OPI API as it
+gains more adoption/support. On the other hand, supporting a custom programmable compression algorithm is likely goin to be
+vendor specific feature.
+- The boundaries on generic vs vendor specific extensions may look blur - there are some easy ones to fall on either side i.e.
+generic vs. vendor specific. The ones in the middle ground can start as vendor specific extensions making their way into
+OPI APIs over time.
 
-All vendor extensions are transparent i.e. not an opaque binary blob of data that is carried in OPI APIs. Making all extensions public
-allows the discussion on the use cases and if they prove to be generic, and include them natively in the OPI APIs.
+All vendor extensions are encouraged to be transparent i.e. not an opaque binary blob of data that is carried in OPI APIs.
+Making all extensions public allows the discussion on the use cases and if they prove to be generic, and include them
+natively in the OPI APIs. However if for any reason to hide the extension in the OPI repo, one can use an opaque `bytes[]`
+protobuf field or use [any](https://developers.google.com/protocol-buffers/docs/proto3#any) construct to achieve that in
+a vendor specific message or a field.
 
 ## Extending OPI Defined Messages
 
-OPI API objects can be extended by adding `oneof` clause to an existing OPI [grpc proto3 message](https://developers.google.com/protocol-buffers/docs/proto3#any). The example below illustrate the extension of a message
+OPI API objects can be extended by adding `oneof` clause to an existing OPI [grpc proto3 message](https://developers.google.com/protocol-buffers/docs/proto3#oneof). The example below illustrate the extension of a message
 and use of the extension in go client library.
 
 ### Defining the extension
@@ -47,22 +59,22 @@ index 1e842e6..d5aab77 100755
 +
 +// NVMeSubsystemSpec AMD Extensions
 +message NVMeSubsystemSpecAMD {
-+    string new_feature  = 1;
++    string new_amd_feature  = 1;
 +}
 +
 +// NVMeSubsystemSpec Intel Extensions
 +message NVMeSubsystemSpecIntel {
-+    string new_feature  = 1;
++    string new_intel_feature  = 1;
 +}
 +
 +// NVMeSubsystemSpec Marvell Extensions
 +message NVMeSubsystemSpecMarvell {
-+    string new_feature  = 1;
++    string new_marvell_feature  = 1;
 +}
 +
 +// NVMeSubsystemSpec Nvidia Extensions
 +message NVMeSubsystemSpecNvidia {
-+    string new_feature  = 1;
++    string new_nvidia_feature  = 1;
  }
 ```
 
@@ -92,7 +104,7 @@ index e9a0c2a..04f6888 100644
                 Subsystem: &pb.NVMeSubsystem{
                         Spec: &pb.NVMeSubsystemSpec{
                                 Id:  &pbc.ObjectKey{Value: "subsystem-test"},
-+                               VendorExtensions: &pb.NVMeSubsystemSpec_Amd{&pb.NVMeSubsystemSpecAMD{NewFeature: "foo"}},
++                               VendorExtensions: &pb.NVMeSubsystemSpec_Amd{&pb.NVMeSubsystemSpecAMD{NewAmdFeature: "foo"}},
                                 Nqn: "nqn.2022-09.io.spdk:opi3"}}})
         if err != nil {
                 log.Fatalf("could not create NVMe subsystem: %v", err)
@@ -110,7 +122,7 @@ index af514a3..0ac317d 100644
                 AllowAnyHost:  true,
                 MaxNamespaces: int(in.Subsystem.Spec.MaxNamespaces),
         }
-+       log.Printf("found vendor extensions = %s", in.Subsystem.Spec.GetAmd().NewFeature)
++       log.Printf("found vendor extensions = %s", in.Subsystem.Spec.GetAmd().NewAmdFeature)
         var result NvmfCreateSubsystemResult
         err := call("nvmf_create_subsystem", &params, &result)
         if err != nil {
--- a/doc/api-extensions.md
+++ b/doc/api-extensions.md
@@ -1,0 +1,233 @@
+# Extending OPI APIs
+
+OPI GRPC APIs may be extended to allow an inclusive way to evolve and extend OPI APIs. The use cases for doing vendor extensions
+may vary, following guidelines can help determine if a new field/message should be considered for an extension as opposed to a
+native OPI field/message.
+
+- Vendor specific feature: If the proposed extension implements a vendor specific feature, then it should be added as an extension.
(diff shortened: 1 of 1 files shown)


doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
+a vendor or generically with a vendor
+- Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general consensus
+that this is a generic feature
+
+All vendor extensions are transparent i.e. not an opaque binary blob of data that is carried in OPI APIs. Making all extensions public
+allows the discussion on the use cases and if they prove to be generic, and include them natively in the OPI APIs.
+
+## Extending OPI Defined Messages
+
+OPI API objects can be extended by adding `oneof` clause to an existing OPI [grpc proto3 message](https://developers.google.com/protocol-buffers/docs/proto3#any). The example below illustrate the extension of a message
```
glimchb, line 22: https://developers.google.com/protocol-buffers/docs/proto3#oneof instead of https://developers.google.com/protocol-buffers/docs/proto3#any right, @jainvipin ?

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
++}
++
++// NVMeSubsystemSpec Intel Extensions
++message NVMeSubsystemSpecIntel {
++    string new_feature  = 1;
++}
++
++// NVMeSubsystemSpec Marvell Extensions
++message NVMeSubsystemSpecMarvell {
++    string new_feature  = 1;
```
glimchb, line 60: do we enforce same type in the vendor extensions ? can it be string vs int between AMD and Nvidia, for example, on the same `new_feature` ?

glimchb (changes requested):
It's an excellent start, few minor nits inside the comments

and one question: another common use case is vendors create their own middle end services, probably... how that would look like ?

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
+# Extending OPI APIs
+
+OPI GRPC APIs may be extended to allow an inclusive way to evolve and extend OPI APIs. The use cases for doing vendor extensions
+may vary, following guidelines can help determine if a new field/message should be considered for an extension as opposed to a
+native OPI field/message.
+
+- Vendor specific feature: If the proposed extension implements a vendor specific feature, then it should be added as an extension.
+A generic feature should be considered as part of the native OPI API in the respective areas (network, storage, storage, etc.).
```
mkalderon, line 8: storage appears twice in brackets

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
+Note that all OPI API fields/messages may not be mandatory, allowing vendors to implement them as their product and/or software
+release process occurs
+- Vendor specific limitation: If the proposed extension exposes a limitation of a product, then it should be added as an extension.
+Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
+a vendor or generically with a vendor
```
mkalderon, line 13: Vipin, can you give an example of limitation vs. modifying generic api to optional ?

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
++
++    // vendor extensions
++    oneof vendor_extensions {
++       NVMeSubsystemSpecAMD amd         = 1024;
++       NVMeSubsystemSpecIntel intel     = 1025;
++       NVMeSubsystemSpecMarvell marvell = 1026;
++       NVMeSubsystemSpecNvidia nvidia   = 1027;
++    }
++}
++
```
mkalderon, line 47: perhaps we need some way of returning which vendor_extension is valid with some query ?

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
++    google.protobuf.FieldMask update_mask = 2;
++}
++
++message GetNVMeVendorextensionRequest {
++    string name = 1 [
++        (google.api.field_behavior) = REQUIRED,
++        (google.api.resource_reference).type = "opi_api.storage.v1/NVMeVendorextension"
++    ];
+ }
+
```
mkalderon, line 231: not sure I follow this example. How would these messages be used by client/server ?

doc/api-extensions.md
jainvipin: yes the ref needs to be fixed. thanks for the catch!

doc/api-extensions.md
jainvipin: we shouldn't in fact likely they are not going to be the same feature, I used the term `new_feature` to illustrate but those features may mean anything to the vendor. If it is the same feature we should bump it up in the OPI APIs.

doc/api-extensions.md
jainvipin: should have been security, thanks for the catch!

doc/api-extensions.md
jainvipin: sure the difference is subtle. I can add following in the document for illustration:
```
For example, exposing a limitation on the number of VFs a device should expose, is a limitation
of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify
as a generic limitation and doesn't belong in vendor specific extensions. On the other hand, if a
vendor's product (say, an older generation of xPU) has a limitation on supporting VirtIO device 
emulation on the PF/VFs
```

Similarly features can be put in the similar category
```
For example, providing deeper IO telemetry (e.g. IO completion time, read/write failure codes)
seems generic enough, but can start as a vendor extension (in Status part of an object), and can
make into OPI API as it gains more adoption/support. On the other hand, supporting a custom
programmable compression algorithm is likely goin to be vendor specific feature.
```

The boundaries are blur, and there are some easy ones to fall on either side (generic vs. vendor specific). The ones in the middle ground can start as vendor specific extensions making their way into OPI APIs over time.

doc/api-extensions.md
jainvipin: I feel capabilities discussion is in general needed for OPI - assuming we are going to have optional features. Having said that backend can easily check if vendor specific option is present using `Is...` generated APIs.

doc/api-extensions.md
jainvipin: I didn't include the client/server example, because they'd look like filling in any standard grpc message in the generated code.

doc/api-extensions.md
glimchb: can you please change the example to use different types and different names just to illustrate that ?

doc/api-extensions.md
glimchb: I like this explanation

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
+Note that all OPI API fields/messages may not be mandatory, allowing vendors to implement them as their product and/or software
+release process occurs
+- Vendor specific limitation: If the proposed extension exposes a limitation of a product, then it should be added as an extension.
+Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
+a vendor or generically with a vendor
+- Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general consensus
+that this is a generic feature
+
+All vendor extensions are transparent i.e. not an opaque binary blob of data that is carried in OPI APIs. Making all extensions public
+allows the discussion on the use cases and if they prove to be generic, and include them natively in the OPI APIs.
```
benlwalker, line 18: I ran this requirement by some large potential consumers of OPI APIs that will have vendor-specific extensions and this requirement is a non-starter. They have extensions that they do not want to make public (and really are specific to them). I've been informed that if this is a requirement of OPI, they'll just skip OPI.

doc/api-extensions.md
```diff
@@ -0,0 +1,233 @@
...
+index 1e842e6..d5aab77 100755
+--- a/storage/v1alpha1/frontend_nvme_pcie.proto
++++ b/storage/v1alpha1/frontend_nvme_pcie.proto
+@@ -148,6 +148,34 @@ message NVMeSubsystemSpec {
+ 
+     // maximum namespaces within a subsystem
+     int64 max_namespaces = 5;
++
++    // vendor extensions
++    oneof vendor_extensions {
```
benlwalker, line 40: What if the extensions are different for every product?

benlwalker (changes requested):

doc/api-extensions.md
jainvipin: they are expected to be, I have modified the examples to illustrate that.

doc/api-extensions.md
jainvipin: well, one can encode binary blobs using, I have changed the language to 'encouraged' vs mandatory. Please review if that would work for you.
@glimchb - pls also review this updated section

doc/api-extensions.md
jainvipin: added

doc/api-extensions.md
jainvipin: fixed

doc/api-extensions.md
jainvipin: fixed

doc/api-extensions.md
jainvipin: fixed

glimchb (approved):
I'm ok with this version

doc/api-extensions.md
glimchb: looks good

benlwalker: What exactly does supporting a blob extension look like? Is there an any inside every oneof?

doc/api-extensions.md
benlwalker: What is the extension is for a user rather than a particular product or vendor? Remember, on most of these products the customer can load their own software.

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
++    string new_nvidia_feature  = 1;
+ }
+```
+
+For improved backward compatibility of the OPI APIs
+
+- Vendor specific messages are enumerated from the values `1024` onwards
+- Specific fields are never included directly in the extensions, but a vendor specific messsage is include
+- Vendor specific message can have its own (and evolving) fields
+- The vendor speific messages are kept in the vendor specific directories, however in the above example, they are kept in the same
```
pathreya, line 86: specific

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
+```diff
+diff --git a/client/frontend.go b/client/frontend.go
+index e9a0c2a..04f6888 100644
+--- a/client/frontend.go
++++ b/client/frontend.go
+@@ -333,6 +333,7 @@ func executeNVMeSubsystem(ctx context.Context, conn grpc.ClientConnInterface) er
+                Subsystem: &pb.NVMeSubsystem{
+                        Spec: &pb.NVMeSubsystemSpec{
+                                Id:  &pbc.ObjectKey{Value: "subsystem-test"},
++                               VendorExtensions: &pb.NVMeSubsystemSpec_Amd{&pb.NVMeSubsystemSpecAMD{NewAmdFeature: "foo"}},
```
pathreya, line 107: It bothers me that the identifiers used here dont match the example above. As in NVMeSubsystemSpec_Amd and NewAmdFeature are not in the example above. Is this as intended?

doc/api-extensions.md
jainvipin: ok so you'd defined a new message or a field as defined in this document that will have a field `bytes[] hidden_feature1` instead of `string new_feature` in the above example - then contents of that field can be serialized and stored and deserialized and restored from the opaque field.

doc/api-extensions.md
benlwalker: But what if they can't wait until OPI adds it and does a release? What I'm getting at is this strategy doesn't work for the users of OPI, and those potential users have made that clear. What does work is a single any field called 'vendor_extension'.

doc/api-extensions.md
jainvipin: that's due to how code is generated, generated code is camel cased from NVMeSubsystemSpec defined in the proto definitions. The important part is that what client/frontend.go use is same as server/frontend.go. I have tried this out in spdk-bridge code to confirm.

doc/api-extensions.md
jainvipin: fixed. thanks

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
+Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
+a vendor or generically with a vendor. For example, exposing a limitation on the number of VFs a device should expose, is a limitation
+of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify as a generic limitation and doesn't
+belong in vendor specific extensions. On the other hand, if a vendor's product (say, an older generation of xPU) has a limitation on
+supporting VirtIO device emulation onf PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
+- Experimental feature: If the proposed extension is an experimental feature that a vendor wants to expose but it lacks general
+consensus that this is a generic feature. For example, providing deeper IO telemetry (e.g. IO completion time, read/write failure
+codes) seems generic enough, but can start as a vendor extension (in Status part of an object), and can make into OPI API as it
+gains more adoption/support. On the other hand, supporting a custom programmable compression algorithm is likely goin to be
```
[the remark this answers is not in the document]
jainvipin: fixed. thanks

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
+A generic feature should be considered as part of the native OPI API in the respective areas (network, storage, security, etc.).
+Note that all OPI API fields/messages may not be mandatory, allowing vendors to implement them as their product and/or software
+release process occurs
+- Vendor specific limitation: If the proposed extension exposes a limitation of a product, then it should be added as an extension.
+Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
+a vendor or generically with a vendor. For example, exposing a limitation on the number of VFs a device should expose, is a limitation
+of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify as a generic limitation and doesn't
+belong in vendor specific extensions. On the other hand, if a vendor's product (say, an older generation of xPU) has a limitation on
+supporting VirtIO device emulation onf PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
```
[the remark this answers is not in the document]
jainvipin: fixed. thanks

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
+A generic feature should be considered as part of the native OPI API in the respective areas (network, storage, security, etc.).
+Note that all OPI API fields/messages may not be mandatory, allowing vendors to implement them as their product and/or software
+release process occurs
+- Vendor specific limitation: If the proposed extension exposes a limitation of a product, then it should be added as an extension.
+Having limits on many features and scale limits may be generic, however a vendor specific limitation is specific to a product from
+a vendor or generically with a vendor. For example, exposing a limitation on the number of VFs a device should expose, is a limitation
+of a device. But all xPUs  limit the maximum number of VFs supported. So this limit may qualify as a generic limitation and doesn't
+belong in vendor specific extensions. On the other hand, if a vendor's product (say, an older generation of xPU) has a limitation on
+supporting VirtIO device emulation on PF/VFs, then that would be a vendor (specifically vendor's product) specific limitation.
```
seroyer, line 16: This example to me feels more like it should be a capabilities exchange based thing, than a vendor specific extension.  But maybe I'm misunderstanding the example?

doc/api-extensions.md
```diff
@@ -0,0 +1,245 @@
...
++
++// NVMeSubsystemSpec Nvidia Extensions
++message NVMeSubsystemSpecNvidia {
++    string new_nvidia_feature  = 1;
+ }
+```
+
+For improved backward compatibility of the OPI APIs
+
+- Vendor specific messages are enumerated from the values `1024` onwards
```
seroyer, line 83: Why `1024`?  Is this a "640K ought to be enough for anybody" kind of situation?  Should a cap on the max value be placed to leave room for another range of values to be added later?

jainvipin: @seroyer - I picked some high number to avoid any conflcit with the field values being added for OPI fields themselves. But this doesn't mean that if we want to go beyond OPI fields can't be added beyond 1024 in a message, it is just that they would not be contiguous (just the optics part of it).

doc/api-extensions.md
jainvipin: To me, limitations and capabilities are the same thing i.e. not having a capability is a limitation. Capabilities (or limitations) could be generic and or they could be vendor specific capabilities (or limitations). I was thinking that this example is really a vendor specific capability and not a generic capability. Maybe the example is not capturing the vendor specific capability as crisply. How about this example: say a vendor has a limitation (i.e. not have a capability) to do encryption only once at line rate - on both inner IP packet (before tunneling) and outer IP packet (post tunneling). So in order to avoid this, it ma expose this limitation by asking user whether to encrypt inner IP header or outer IP header when implementing IP layer encryption. Of course use can choose to do encryption at both layers, but then it wouldn't be giving them the line rate performance. So exposing an extension to the user to overcome this limitation may be vendor specific.

glimchb:
PR has enough comments already and I think good start to merge at this point
Every comments from everybody are noted, please feel free to create a new PR so we can have iterative development going here
To scale we need other people start proposing pull requests as well

doc/api-extensions.md
seroyer: This new example makes more sense for me, thanks.

doc/api-extensions.md
seroyer: I guess I'm suggesting we explicitly document the vendor extension range.  For example 1024-2047 and 2048+ are reserved for future use or something.

jainvipin: makes sense @seroyer - there is a way to do that natively in proto definitions. may be you and I can submit a PR for that separately. Pls let me know.

doc/api-extensions.md
jainvipin: I will include that in updated text in subsequent PR; thanks you @seroyer for your input!

PR #230, opened 2023-01-05
[network] auto-generate protos

glimchb:


sandersms:
Why are you removing the v1?  Wouldn't this be the proper place for a common set of API definitions or is that now going to be the opinetcommon?  Shouldn't any of the defined api's that we have done in the network v1 be moved to the opinetcommon folder for now until we decide that they can be deleted, updated, depicated?

glimchb:
> Why are you removing the v1? Wouldn't this be the proper place for a common set of API definitions or is that now going to be the opinetcommon? 

Yes, https://github.com/opiproject/opi-api/tree/main/network/opinetcommon should be the new place for all common network APIs

> Shouldn't any of the defined api's that we have done in the network v1 be moved to the opinetcommon folder for now until we decide that they can be deleted, updated, depicated?

I see only 1 proto file defined today https://github.com/opiproject/opi-api/blob/main/network/v1/oc_interfaces.proto
I can move to `opinetcommon` for example...

sandersms (approved):
Looks good to me with the oc_interfaces.proto being moved over as a mapping of the openconfig interfaces model.

seroyer (approved):

PR #234, opened 2023-01-18
first revision - cloud protos

jainvipin:
- Please go over README.md for any comments about what certain objects mean, and why they are relevant
- all protos are verified to compile and autogenerate files
- avoiding autogeneration to reduce clutter and easier review
- haven't defined the RPC messages, will do once proto feedback is incorporated
- tried to split the code between opnnetcommon (something that may be useful elsewhere) from cloud apis
- added documentation for the cloud protos to help understand the object model

ping @glimchb @sandersms @pathreya @mestery


glimchb (commented):
lgtm. good first revision

sandersms (approved):
Looks good.
The combination of the Readme and the device.proto comments clearly define the layered approach related to the cloud protos.  The device to interface to higher levels flowing up is clear.

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+//  rpc SecurityRuleGet (SecurityRuleGetRequest) returns (SecurityRuleGetResponse) {}
+//  rpc SecurityRuleDelete (SecurityRuleDeleteRequest) returns (SecurityRuleDeleteResponse) {}
+//
+//  rpc SecurityProfileCreate (SecurityProfileRequest) returns (SecurityProfileResponse) {}
+//  rpc SecurityProfileUpdate (SecurityProfileRequest) returns (SecurityProfileResponse) {}
+//  rpc SecurityProfileGet (SecurityProfileGetRequest) returns (SecurityProfileGetResponse) {}
+//  rpc SecurityProfileDelete (SecurityProfileDeleteRequest) returns (SecurityProfileDeleteResponse) {}
+//
+//  rpc SecurityPolicyLookup (SecurityPolicyLookupRequest) returns (SecurityPolicyLookupResponse) {}
+//
```
timmichels, line 128: Why all the commented out rpc calls?

network/cloud/v1alpha1/interface.proto
```diff
@@ -0,0 +1,195 @@
...
+  repeated BGPCommunity communities = 2;
+}
+
+// Inband control interface configuration
+message ControlIfSpec {
+  // IP address hosted on the inband control interface
+  network.opinetcommon.v1alpha1.IPPrefix prefix = 1;
+  // mac address of the interface
+  bytes mac_address = 2;
+}
```
timmichels, line 105: Where is the inband control interface terminated on the host - the hypervisor, the VNIC, somewhere else?

network/cloud/v1alpha1/interface.proto
```diff
@@ -0,0 +1,195 @@
...
+  IF_TYPE_UPLINK_PC   = 2;
+  // L3 interface
+  IF_TYPE_L3          = 3;
+  // loopback interface
+  IF_TYPE_LOOPBACK    = 4;
+  // inband management/control interface
+  IF_TYPE_CONTROL     = 5;
+  // host visible PF/VF device
+  IF_TYPE_HOST        = 6;
+}
```
timmichels, line 183: In Interface types, how is Uplink_PC different than just Uplink?

network/cloud/v1alpha1/networkpolicy.proto
```diff
@@ -0,0 +1,371 @@
+// Copyright (c) 2023 Pensando, AMD Inc, or its subsidiaries.
+// protobuf specification for network security policies
+
```
timmichels, line 3: The Network Policy API below to me hints at an underlying HW implementation and architecture in its representations and format.  My concern is that it will only map to the xPU it was originally created for.  Do other xPU vendors believe they can consume this API?

timmichels (changes requested):
My comment about representations captured herein being too specific to a single xPU architecutre is my largest concern.  Many of the elements herein are fine, but some perhaps are not generalized enough and assume a specific underlying match/action HW pipeline.

network/cloud/README.md
```diff
@@ -1 +1,152 @@
...
+                                            \
+                                             \
+                                            To Internet (via gateway devices)
+
+```
+
+Each workload (VM, BM or a container) is identified by the xPU as a vNIC (virtual NIC).
+A VNIC could be a physical pcie device (PF/VF) or a virtual identifier (e.g. a VLAN-id
+tagged by the hypervisor that runs on the host. Once xPU identifes the workload, it then
+maps its traffic into a relevant VPC. Of course, the
```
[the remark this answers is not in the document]
jainvipin: yes a line got deleted; fixed.

network/cloud/v1alpha1/networkpolicy.proto
```diff
@@ -0,0 +1,371 @@
...
+  ALG_TYPE_DNS         = 3;
+  // SUNRPC ALG
+  ALG_TYPE_SUNRPC      = 4;
+  // MSRPC ALG
+  ALG_TYPE_MSRPC       = 5;
+  // RTSP ALG
+  ALG_TYPE_RTSP        = 6;
+}
+
+// PolicyRuleFormat captures what format is used for the rules in a given
```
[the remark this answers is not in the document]
jainvipin: RuleFormat (flattened vs. compact vs. none) - actually makes the policy implementation independent. It allows xPUs that have flat flow table to work as flattened rules, whereas richer implementations can implement aggregate rules in the datapath pipeline (e.g. programmable P4 pipeline can handle ranges, lists in the datapath), and the caller can take advantage of it. IF you still think otherwise, I will remove RuleFormat field.

network/cloud/v1alpha1/subnet.proto
```diff
@@ -0,0 +1,110 @@
...
+  //   (so vnics can be configured with VnicEncap as ENCAP_TYPE_NONE, so they
+  //   inherit the subnet's encap)
+  // - it is also possible to have subnet encap as ENCAP_TYPE_DOT1Q with
+  //   certain tag value but some vnic's in that subnet to have different
+  //   DOT1Q encap than that of subnet
+  network.opinetcommon.v1alpha1.Encap access_encap = 14;
+  // encapped traffic coming with the specified encap from fabric side will be
+  // mapped to this subnet and traffic from workloads in this subnet will carry
+  // this encap towards the fabric side
+  network.opinetcommon.v1alpha1.Encap fabric_encap = 15;
```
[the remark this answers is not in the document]
jainvipin: fabric encap is a multiplexed way to specifying encapsulation type (mpls, nvgre, vxlan, etc.) and encap value (e.g. VNI, mpls-label, etc.). So by having an encap for a subnet allows a VNI to be assigned directly to a subnet (as per the comment). Happy to expand more in the comment of this field.

network/cloud/README.md
```diff
@@ -1 +1,152 @@
...
+                                     |           v                                          |
+                                     |       [Send out on uplink port to Network Switch]    |
+                                     |           |                                          |
+  Port object (physical port)--------+---------> |                                          |
+                                     |           |                                          |
+                                     .___________|__________________________________________.
+                                                 |
+                                                 v
+                                          To Underlay Network Switch
+
```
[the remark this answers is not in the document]
jainvipin: yes, first one is to specify what it logically represents, second one specifies how it gets used in the pipeline for VPC identification.

network/cloud/v1alpha1/device.proto
jainvipin: But after a discussion if we need one service with all APIs or different grpc services. Any opinons? I'd prefer one grpc service, but wasn't sure what others would prefer.I could take ti up in this PR, or another PR.

network/cloud/v1alpha1/interface.proto
jainvipin: Uplink PC is a port-channel i.e. multiple ports can form a 802.3ad link aggregation port (kinda like NIC bonding on linux) towards the Top of rack from the xPU's ports.

network/cloud/v1alpha1/networkpolicy.proto
jainvipin: Let's discuss - would love opinion from others.
The structure of policy is very generic SecurityPolicy has a bunch of ordered Rules, each rule has priority(order), matching criteria (5 tuple in various forms), and action (permit/deny). Besides it can be applied in ingress or egress direciton at various attachment points (subnet, vpc, etc.). Do you think, it would help if this hierarchy of network security policy is documented?

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  // system MAC address (unique for a device in cluster)
+  bytes system_mac_address = 2;
+  // vendor identifier
+  string vendor_id = 3;
+  // chip type
+  string chip_type = 4;
+  // OS version
+  string os_version = 5;
+  // number of PCIe ports
+  int32 pcie_port_count = 6;
```
pathreya, line 206: What is pcie_port logically? Is it different than the pf that the host sees?

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  // Vlan ID (0-4095) is the management network's vlan tag (zero = untagged)
+  int32 vlan = 3;
+  // IP address of the default gateway in the management vlan
+  network.opinetcommon.v1alpha1.IPAddress gateway_ip = 4;
+  // MAC address of the default gateway in the management vlan
+  bytes gateway_mac = 5;
+}
+
+// device's network status
+message DeviceStatus {
```
pathreya, line 194: Is this message represents something the device natively advertises(like info and caps) OR is it the status after a configuration? If its latter, i think a structure is needed for capabilities and immutable device information which can be separate from the status fields which are dynamic based on configuration or state of the device.

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  // underlay ipv6 address for tunnel initiation/termination
+  network.opinetcommon.v1alpha1.IPAddress ipv6_address = 3;
+  // device's mac address - for outer ethernet header
+  bytes mac_addr = 4;
+  // IP address of the underlay gateway in the underlay
+  network.opinetcommon.v1alpha1.IPAddress gateway_ip = 5;
+  // number and type of pcie functions exposed
+  PCIeFunctionsSpec pcie_functions = 6;
+  // ovelay routing
+  bool overlay_routing_enabled = 7;
```
pathreya, line 164: I think it affects the pipeline? Perhaps not a global device config then.

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  common.v1.ObjectKey id = 1;
+
+  // underlay ipv4 address for tunnel initiation/termination
+  network.opinetcommon.v1alpha1.IPAddress ipv4_address = 2;
+  // underlay ipv6 address for tunnel initiation/termination
+  network.opinetcommon.v1alpha1.IPAddress ipv6_address = 3;
+  // device's mac address - for outer ethernet header
+  bytes mac_addr = 4;
+  // IP address of the underlay gateway in the underlay
+  network.opinetcommon.v1alpha1.IPAddress gateway_ip = 5;
```
pathreya, line 160: Is this(and ipvX_address) best addressed by API Spec OR is it better configured by the lifecycle management?
@glimchb thoughts?

pathreya (changes requested):

network/cloud/v1alpha1/device.proto
glimchb: I favor the approach to comment out RPC's in this PR and focus on messages and if all agree, do the RPC pull request separately following this one...

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+// critical alerts effecting the health of the system
+message SystemAlert {
+  // alert timestamp
+  google.protobuf.Timestamp alert_time  = 1;
+  // name of the alert
+  string                    alert_description = 2;
+}
+
+// Device Create Request
+message CreateDeviceRequest {
```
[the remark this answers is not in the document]
glimchb: - from Controller to xPU
- Yes, lifecycle management group can investigate how to advertise management end points so Controllers can find xPUs. It could be static config or it could be dynamic advertisement... ssdp... dhcp... mdns... slaac... In any case should be out of scope for this PR

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  // ovelay routing
+  bool overlay_routing_enabled = 7; 
+  // system name is used as named identifier in protocols (e.g. LLDP)
+  string systemname = 8;
+  // management network details
+  MgmtNetworkSpec mgmt_network_spec = 9;
+}
+
+// PCIe functions to be configured on the device
+message PCIeFunctionsSpec {
```
glimchb, line 172: we can potentially use `PciEndpoint` from https://github.com/opiproject/opi-api/blob/main/storage/v1alpha1/opicommon.proto#L22

network/cloud/v1alpha1/networkpolicy.proto
```diff
@@ -0,0 +1,371 @@
+// Copyright (c) 2023 Pensando, AMD Inc, or its subsidiaries.
```
pathreya, line 1: I think we can avoid NetworkPolicy for now. I wont be reviewing it in this PR.

network/cloud/v1alpha1/port.proto
```diff
@@ -0,0 +1,369 @@
...
+// port status
+message PortStatus {
+  // encoded interface index
+  // (-- api-linter: core::0141::forbidden-types=disabled
+  //     aip.dev/not-precedent: interface index is an opaque uint32 value. --)
+  uint32                     if_index    = 1;
+  // operational link status
+  PortLinkStatus             link_status = 2;
+  // transceiver status
+  PortXcvrStatus             xcvr_status = 3;
```
pathreya, line 66: I'd like xcvr_status, fsm_xxx to be optional.

network/cloud/v1alpha1/port.proto
```diff
@@ -0,0 +1,369 @@
...
+  // internal port link state machine
+  PortLinkFSM                fsm_state   = 4;
+  // mac associated with the port
+  int32                      mac_id      = 5;
+  // mac channel associated with the port
+  int32                      mac_ch      = 6;
```
pathreya, line 72: Can you clarify what is mac channel?

network/cloud/v1alpha1/cproute.proto
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2023 Pensando, AMD Inc, or its subsidiaries.
+// protobuf specification for Control Plane (i.e. underlay) Route Table
```
pathreya, line 2: I think CP is a misnomer since this route table is consulted for datapath of all traffic(tenant or xPU). Perhaps ulroute is more apt.

network/cloud/v1alpha1/vnic.proto
```diff
@@ -0,0 +1,124 @@
...
+  // enable or disable reverse path checks while rx/tx traffic from/to this vnic
+  bool  source_guard_enable = 5;
+  // fabric encap information to be used for traffic originated from this vnic
+  network.opinetcommon.v1alpha1.Encap fabric_encap = 6;
+  // VNFs (virtual network functions are workloads that implement network
+  // features such as firewalls, virtual switches/bridges, security monitoring, etc.
+  // since VNFs are devices in the middle, they can rx/tx traffic on behalf of other
+  // workloads to allow for source/destination checks, and exceptions that of a mapping
+  bool vnf = 7;
+  // identifier of ingress IPv4 security policy to be enforced
```
pathreya, line 46: fields related to policy need to be optional.

network/cloud/v1alpha1/vnic.proto
```diff
@@ -0,0 +1,124 @@
...
+  // max_sessions, if set, is total number of active sessions (across all
+  // protocols) allowed from/to this vnic; zero means unlimited
+  int32  max_sessions = 20;
+  // guest workload's MAC in rx/tx direction is rewritten with this mac if non zero
+  bytes public_mac_address = 25;
+  // if allow_internet_access is set to false and traffic from the vnic hits a
+  // route of class ROUTE_CLASS_INTERNET, then traffic is dropped. To allow
+  // Internet connectivity for the vnic, AllowInternetAccess must be set to
+  // true
+  bool allow_internet_access   = 28;
```
pathreya, line 70: I think this setting belongs under network policy

network/cloud/v1alpha1/vnic.proto
```diff
@@ -0,0 +1,124 @@
...
+  // max_cps, if non-zero, is the maximum no. of connections per second (cps)
+  // allowed for this vnic and if cps exceeds this configured limit all
+  // new connections will get dropped; zero means unlimited
+  int32 max_cps = 29;
```
pathreya, line 74: CPS settings belong under network policy.

network/cloud/v1alpha1/vnic.proto
```diff
@@ -0,0 +1,124 @@
...
+  common.v1.ObjectKey v4_route_table_id = 33;
+  // identifier of the IPv6 route table to be used, if any
+  common.v1.ObjectKey  v6_route_table_id = 34;
+  // vnic if assigned a VIP in the underlay, can be used as
+  // outer encap source IP, if configured, for traffic going to certain vpc
+  // private service endpoints
+  network.opinetcommon.v1alpha1.IPAddress service_ip = 36;
+  // max_tcp_sessions is the maximum number of TCP sessions allowed from/to this
+  // vnic, if it is non-zero; zero implies no limit
+  int32 max_tcp_sessions = 38;
```
pathreya, line 92: max_xxx belong under policy

network/cloud/README.md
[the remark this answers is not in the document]
jainvipin: First thing to be done for egress traffic is always tenant/vpc identification; usually cloud provider wouldn't trust the IP address (from a mapping object) to identify the workload. The trusted way to identify the workload is to use its VNIC. Once the VPC is identified, it tells various attributes of the VPC to help route the traffic (besides other things). Mapping object at this stage can be used to to source-guard checks i.e. ensure that sender is not spoofing the src-ip/mac. Later mapping object is used to do the routing lookup because mapping is also a /32 or /128 route. identifying the destination host.

network/cloud/v1alpha1/device.proto
jainvipin: ack @glimchb - incremental progress is better, to avoid discussion diluation.

network/cloud/v1alpha1/device.proto
```diff
@@ -0,0 +1,246 @@
...
+  PCIeFunctionsSpec pcie_functions = 6;
+  // ovelay routing
+  bool overlay_routing_enabled = 7; 
+  // system name is used as named identifier in protocols (e.g. LLDP)
+  string systemname = 8;
+  // management network details
+  MgmtNetworkSpec mgmt_network_spec = 9;
+}
+
+// PCIe functions to be configured on the device
```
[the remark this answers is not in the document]
jainvipin: yes the VFs/PFs would be visible on the host. For virtualization, these VFs can be mapped into VMs (with hypervisor bypass), or inside a K8s pod. OTOH, if it is a bare-metal workload these PFs/VFs are used by the tenant. e.g. bring your own hypervisor is a good use case for scheduling BM resources in public cloud.

network/cloud/v1alpha1/device.proto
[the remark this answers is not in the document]
jainvipin: @glimchb is correct; all GRPC APIs are from controller to xPU. xPU is managed northbound by cloud infra controller. We need to tee off from lifecycle APIs to be handed over to network SDN control plane to start managing the datapath. Clearly no tenant workoads can be provisioned while all this is being bootstrapped.

I have updated README.md with a section describing the management path and OS boot dependencies.

network/cloud/v1alpha1/device.proto
jainvipin: no it is number of PFs a cloud provider seeks to be exposed from this device.

network/cloud/v1alpha1/device.proto
jainvipin: good point - I wasn't sure if capabilities belonged to network APIs, but I think we can put network capabilities in network APIs. Given that,  I have added another API to fetch the device capabilities. I have added a few candidate fields in there, but feel free to submit a PR to add more capabilities there. Pls see `DeviceCapabilities` in the updated PR. Also note @timmichels, this would also address the concern about flattened vs. compact rules comment you made earlier.

network/cloud/v1alpha1/vnic.proto
jainvipin: this is not underlay configuration - this is a limitation put on to a tenant's workload (via VNIC object) by the cloud provider to help manage the connection/sec resources fairly among various tenants.

network/cloud/v1alpha1/vnic.proto
jainvipin: Well the thought was that network security policy would defined permit/deny rules based on the 5-tuple, and also network security policy is defined by the customer to restrict who can talk to who in tenant network. However even if a VNIC/VPC/Subnet is not configured with any security policy a cloud infra may want to limit the max# sessions based on the workload profile (e.g. an instance that is higher #CPUs may be allowed more sessions vs. a tiny instance may be restricted).

network/cloud/v1alpha1/vnic.proto
jainvipin: yes we haven't annotated any of the fiels as mandatory vs. optional. But clearly security policy on VNIC is optional and is usually triggered by the tenant's intent.

network/cloud/v1alpha1/vnic.proto
jainvipin: Permitting internet access is typically a workload instance (e.g. VM instance) attribute. Having said that yes this is a policy but for a workload (i.e. a VNIC). If we move this to network security policy we'd have to define the behavior for other entities that can also be attaching network security policy.
Let's hear from others about how to define workload specific attributes in a generic network policy structure. Or, would it help if we take all the policy specific fields in VNIC in a separate message and refer to them here, instead of calling out the fields natively?

network/cloud/v1alpha1/port.proto
jainvipin: ack

network/cloud/v1alpha1/networkpolicy.proto
jainvipin: ok - let's hear from others. I can take it out. However this was targeted to be kept in the first rev. @glimchb @sandersms ?

network/cloud/v1alpha1/bgp.proto
```diff
@@ -0,0 +1,686 @@
+// Copyright (c) 2023 Pensando, AMD Inc, or its subsidiaries.
```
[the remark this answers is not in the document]
jainvipin: ack - pls check the capability message. I can also remove the bgp.proto in this PR and come to it later if there is a consensus on moving it out for now.

network/cloud/v1alpha1/cproute.proto
jainvipin: sure - let's find a better name. I'd prefer `UnderlayRoute` in that case. If others want to opine, this would be a good time.

network/cloud/v1alpha1/device.proto
jainvipin: It affects the pipeline but it is a networking configuration so we need to define it at some place. If this field set then overlay traffic is routed, otherwise it is bridged i.e. forwarded within the subnet only based on the MAC address. Can be useful for EVPN use case that someone brought up. If using bridging (e.g. EVPN) it make sense to do it for all VPCs i.e. globally and not selectively, otherwise it can get confusing.

network/cloud/v1alpha1/device.proto
jainvipin: IMO, lifecycle should be only relying upon and using management interface's connection. This is the gateway IP configuration for the underlay.

network/cloud/v1alpha1/device.proto
jainvipin: may be I named it poorly - this message is really specifying how many pcie functions to expose, to the host. Where as https://github.com/opiproject/opi-api/blob/main/storage/v1alpha1/opicommon.proto#L22  indicates the actual details the the PCIe endpoint which definitely belongs in there for storage related VFs.

network/cloud/v1alpha1/cproute.proto
```diff
@@ -0,0 +1,84 @@
...
+  // destination address
+  network.opinetcommon.v1alpha1.IPPrefix dest_prefix = 3;
+  // next-hop address
+  network.opinetcommon.v1alpha1.IPAddress next_hop_address = 4;
+  // route is enabled or not
+  network.opinetcommon.v1alpha1.AdminState state = 5;
+  // override admin distance
+  bool override = 6;
+  // admin distance
+  uint32 admin_dist = 7;
```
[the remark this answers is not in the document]
jainvipin: This configuration has is not the metric in linux kernel routes. For static routes (vs dynamic where distance is computed by protocol), this is the configured admin-distance set by the controller for this route to help find the best next hop in RIB. This control can be used by the static routing to configure specific paths (via distance) for underlay routes.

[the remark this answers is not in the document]
jainvipin: I have updated the comment for this field in the PR.

network/cloud/v1alpha1/cproute.proto
```diff
@@ -0,0 +1,84 @@
...
+  // route table id this route belongs to
+  common.v1.ObjectKey route_table_id = 2;
+  // destination address
+  network.opinetcommon.v1alpha1.IPPrefix dest_prefix = 3;
+  // next-hop address
+  network.opinetcommon.v1alpha1.IPAddress next_hop_address = 4;
+  // route is enabled or not
+  network.opinetcommon.v1alpha1.AdminState state = 5;
+  // override admin distance
+  bool override = 6;
```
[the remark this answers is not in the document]
jainvipin: not really, this can be used to override admin distance of a route that is computed equidistant in the default calcuation. Since 'zero' is a valid admin_distance, this bool really idenfies if admin_dist is valid or not.

network/cloud/v1alpha1/device.proto
jainvipin: On second thought, this was called control plane route, because this is a static substitution to what a routing protocol (aka control plane) would have computed, and so control plane route. Either way - I am okay, just providing the context for everyone.

chaudron:
I like the README.md as it explains the use case and bits of the pipeline. 

As mentioned earlier, what I think we should use, are the OpenConfig definitions as they are already in use for a while, and major routing vendors are using them. If we do miss items in the definitions we could add them.

During the January 19th meeting, someone brought up that we need some form of capability checking. Briefly looking at the gNMI protocol and clients, this seems to be built in (https://github.com/akarneliuk/pygnmi, https://www.openconfig.net/docs/gnmi/gnmi-specification/#32-capability-discovery).

I know someone was looking at providing an example client/server, but this might be delayed due to the Chinese New Year.

So my take would be, rather than trying to push our own Protobuf definitions, try to use OpenConfig + gNMI and add any missing mandatory attributes.

Briefly looking at the general interface (.proto), most of the fields are there. If found this page explaining a bit about how OpenConfig definitions should be converted to protocol buff, https://www.openconfig.net/docs/gnmi/protobuf-vals/

//Eelco

jainvipin:
> I like the README.md as it explains the use case and bits of the pipeline.
> 
> As mentioned earlier, what I think we should use, are the OpenConfig definitions as they are already in use for a while, and major routing vendors are using them. If we do miss items in the definitions we could add them.
> 
> During the January 19th meeting, someone brought up that we need some form of capability checking. Briefly looking at the gNMI protocol and clients, this seems to be built in (https://github.com/akarneliuk/pygnmi, https://www.openconfig.net/docs/gnmi/gnmi-specification/#32-capability-discovery).
> 
> I know someone was looking at providing an example client/server, but this might be delayed due to the Chinese New Year.
> 
> So my take would be, rather than trying to push our own Protobuf definitions, try to use OpenConfig + gNMI and add any missing mandatory attributes.
> 
> Briefly looking at the general interface (.proto), most of the fields are there. If found this page explaining a bit about how OpenConfig definitions should be converted to protocol buff, https://www.openconfig.net/docs/gnmi/protobuf-vals/
> 
> //Eelco
thanks @chaudron - I will let others also opine on this. 

I am not a savvy openconfig user (esp when it comes to proto generation) so for me, I wouldn't mind (or it will help) if someone can submit a PR for these definitions in Openconfig that can generate these proto definitions. OTOH, assuming we can generate similar kind of proto definitions out of openconfig YANG model, we can start working on the backend reference implementation around these APIs to make progress.

chaudron:
> thanks @chaudron - I will let others also opine on this.
> 
> I am not a savvy openconfig user (esp when it comes to proto generation) so for me, I wouldn't mind (or it will help) if someone can submit a PR for these definitions in Openconfig that can generate these proto definitions. OTOH, assuming we can generate similar kind of proto definitions out of openconfig YANG model, we can start working on the backend reference implementation around these APIs to make progress.

The same for me, however, we had an internal discussion around the subject of OpenConfig/IETF and we are trying to get some "experts" involved. Give us a couple of days.

network/cloud/v1alpha1/cproute.proto
```diff
@@ -0,0 +1,85 @@
...
+  network.opinetcommon.v1alpha1.IPPrefix dest_prefix = 2;
+  // next-hop address
+  network.opinetcommon.v1alpha1.IPAddress next_hop_address = 3;
+  // interface index, if the next-hop is an interface
+  int32 if_index = 5;
+  // type of route
+  network.opinetcommon.v1alpha1.RouteType type = 6;
+  // protocol via which the route is learned
+  network.opinetcommon.v1alpha1.RouteProtocol proto = 7;
+  // The number of seconds since this route was last updated or otherwise determined to be correct.
```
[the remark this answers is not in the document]
jainvipin: @pathreya - as for making these fields optional, certainly we can add google options to be tagged to these fields.
@glimchb - WDYT? should I include grpc `optional` attributes to various fields in this PR? I can do a sweep at all the fields and do it. There are a large # of fields optional in this API, not just the one @pathreya points out.

network/cloud/v1alpha1/cproute.proto
```diff
@@ -0,0 +1,85 @@
...
+  int32 age = 8;
+  // The primary routing metric for this route
+  int32 metric1 = 9;
```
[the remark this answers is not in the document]
jainvipin: fixed

network/cloud/v1alpha1/cproute.proto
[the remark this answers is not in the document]
jainvipin: rest of the comments incorporated @pathreya - pls review.

pathreya (approved):
Based on offline exchange, I will submit PR for fields that will need to be optional for Marvell to be compliant with this spec. Optional fields will be identified in the API using https://linter.aip.dev/203/optional.
Hence this PR is approved from my end.

glimchb:
@timmichels do you still have an objection or we can merge this ?

glimchb:
good first proposal, we can iterate on top of this. merging

PR #237, opened 2023-01-25
New rpc service to attach/detach a controller to a namespace.

VaibhavJMarvell:
Added attach/detach controller to cater to use-case of an existing customer.

Signed-off-by: Vaibhav Jain <vaibhavj@marvell.com>


glimchb (changes requested):
how this si different from `CreateNVMeNamespace` ?

VaibhavJMarvell:
> how this si different from `CreateNVMeNamespace` ?

> We have an existing customer whose use-case requires attaching/detaching specific controllers to a namespace. In the current scenario of 'CreateNVMeNamespace', all the controllers are attached to the newly created namespace.

glimchb:
> > how this si different from `CreateNVMeNamespace` ?
> 
> We have an existing customer whose use-case requires attaching/detaching specific controllers to a namespace. In the current scenario of 'CreateNVMeNamespace', all the controllers are attached to the newly created namespace.

so maybe just add field `ControllerId` to existing `CreateNVMeNamespace` ? instead of new RPCs?

storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -398,6 +412,13 @@ message NVMeNamespaceStatsResponse {
     VolumeStats stats = 2;
 }
 
+message ControllerNVMeNamespaceRequest {
+    string subsystem = 1 [
+        (google.api.field_behavior) = REQUIRED
+    ];
+    string controller = 2 [(google.api.field_behavior) = REQUIRED];
+    string nv_me_namespace_id = 3 [(google.api.field_behavior) = REQUIRED];
```
jainvipin, line 420: should this be `common.v1.ObjectKey` which bytes all ids are like that. Same for `controller`?

jainvipin (changes requested):
+1 on the need for this API - we also see the need to explicitly enble/disable the namespace within a controller.
a few minor comments inline @VaibhavJMarvell

jainvipin:
> > > how this si different from `CreateNVMeNamespace` ?
> > 
> > 
> > We have an existing customer whose use-case requires attaching/detaching specific controllers to a namespace. In the current scenario of 'CreateNVMeNamespace', all the controllers are attached to the newly created namespace.
> 
> so maybe just add field `ControllerId` to existing `CreateNVMeNamespace` ? instead of new RPCs? see [ca2c25b](https://github.com/opiproject/opi-api/commit/ca2c25b808b32b217e946ec3145863f1e8814f03)

we could do that to, but this API would then become enable/disable to explicitly enable/disable a Namespace on controller.

VaibhavJMarvell (commented):
I have changed the controller and namespace types to ObjectKey.

glimchb:
> > > > how this si different from `CreateNVMeNamespace` ?
> > > 
> > > 
> > > We have an existing customer whose use-case requires attaching/detaching specific controllers to a namespace. In the current scenario of 'CreateNVMeNamespace', all the controllers are attached to the newly created namespace.
> > 
> > 
> > so maybe just add field `ControllerId` to existing `CreateNVMeNamespace` ? instead of new RPCs? see [ca2c25b](https://github.com/opiproject/opi-api/commit/ca2c25b808b32b217e946ec3145863f1e8814f03)
> 
> we could do that to, but this API would then become explicit enable/disable API to toggle namespace on controller.

seems to me like we should use `UpdateNVMeNamespace` if we want to toggle enable/disable bit... instead of new RPC

jainvipin:
> > > > > how this si different from `CreateNVMeNamespace` ?
> > > > 
> > > > 
> > > > We have an existing customer whose use-case requires attaching/detaching specific controllers to a namespace. In the current scenario of 'CreateNVMeNamespace', all the controllers are attached to the newly created namespace.
> > > 
> > > 
> > > so maybe just add field `ControllerId` to existing `CreateNVMeNamespace` ? instead of new RPCs? see [ca2c25b](https://github.com/opiproject/opi-api/commit/ca2c25b808b32b217e946ec3145863f1e8814f03)
> > 
> > 
> > we could do that to, but this API would then become explicit enable/disable API to toggle namespace on controller.
> 
> seems to me like we should use `UpdateNVMeNamespace` if we want to toggle enable/disable bit... instead of new RPC

we need to add enable/disable in NameSpaceSpec if we do it via namespace object; that'd work too, but having an admin know to enable a namespace on a subsystem/controller is the ask here.

glimchb:
> we need to add enable/disable in NameSpaceSpec if we do it via namespace object; that'd work too, but having an admin know to enable a namespace on a subsystem/controller is the ask here.

agree, let's add Enable/Disable to `NameSpaceSpec` and then using `Update` RPC we can toggle it

glimchb:
@VaibhavJMarvell @jainvipin can you please continue this ? it is open for a long time

PR #277, opened 2023-04-18
Fix linter errors in /common folder and inventory

sandersms:
Fix the Linter errors in the inventory protobuf file found when building the /common/Makefile.

Updated the inventory.proto file to include changes to the rpc for google apis as well as the Inventory request and response.
InventoryGet  --> GetInventory
InventoryGetRequest --> GetInventoryRequest
InventoryGetResponse --> Inventory


glimchb (approved):
thanks for fixing this @sanders-mark

PR #279, opened 2023-04-27
Replace plantuml with mermaid

glimchb:
Much easier to maintain and integrated with github


seroyer (approved):
Nice!

PR #285, opened 2023-05-26
move the name to the top level, change references to call by name ref…

jainvipin:
- move name to top level
- move id to name
- change references to _name_ref from _id

Other misc changes to identify the resource object in protobuf


network/cloud/v1alpha1/bgp.proto
```diff
@@ -19,21 +19,25 @@ option java_outer_classname = "BGPProto";
...
+  BgpSpec spec = 2;
   // BGP Status
-  BgpStatus status = 2;
+  BgpStatus status = 3;
 }
 
 // BGP configuration
 message BgpSpec {
-  // unique key/identifier of BGP config
-  common.v1.ObjectKey id = 1;
```
glimchb, line 36: adjust numbering after removing this first field ?

network/cloud/v1alpha1/bgp.proto
```diff
@@ -19,21 +19,25 @@ option java_outer_classname = "BGPProto";
...
 import "networktypes.proto";
+import "google/api/resource.proto";
 
 // BGP object
 message Bgp {
+  option (google.api.resource) = {
+    type: "opi_api.network.cloud.v1alpha1/bgp"
+    pattern: "bgps/{bgp}"
+  };
+  // unique key/identifier of BGP config
```
glimchb, line 31: the comment could be a bit more elaborate like this:
```
    // name is an opaque object handle that is not user settable.
    // name will be returned with created object
```

jainvipin: sure - I can add - but adding this in every object sounded too repetitive, for every object.

network/cloud/v1alpha1/bgp.proto
jainvipin: good catch, I thought I did it for all objects; apparently not!

sandersms (approved):
Looks good with changes incorporated

PR #291, opened 2023-06-01
Fix QoS volume limit names

artek-koltun:
Use `min_limit`/`max_limit` as QoS limit names in the volume, as for devices


glimchb:
@artek-koltun for my understanding, why do you think this is better name ?

artek-koltun:
> @artek-koltun for my understanding, why do you think this is better name ?

It is just to be consistent with the names used in [scsi](https://github.com/opiproject/opi-api/blob/52f6d9b6b6a6118760e9fd657ec984c6bbab4584/storage/v1alpha1/frontend_virtio_scsi.proto#L158), [nvme](https://github.com/opiproject/opi-api/blob/52f6d9b6b6a6118760e9fd657ec984c6bbab4584/storage/v1alpha1/frontend_nvme_pcie.proto#L224), etc.

glimchb (approved):
lgtm

PR #292, opened 2023-06-01
Fix for AdminState redeclared - Issue#289

sandersms:


sandersms:
> thanks for fixing this @sanders-mark

No problem - changes were done last night; but, had to generate the auto files.

PR #293, opened 2023-06-01
[cloud] rename id->name

glimchb:
- [cloud] rename id->name
- [cloud] auto-generate protos

see https://google.aip.dev/135#request-message
see https://google.aip.dev/131#request-message


sandersms (approved):
LGTM

PR #296, opened 2023-06-02
Fix circle import in openconfig import

artek-koltun:
Network openconfig interface had a circle dependency between `github.com/opiproject/opi-api/network/v1alpha1/gen/go` and `github.com/opiproject/opi-api/network/opinetcommon/v1alpha1/gen/go` which prevented from using this Go package


glimchb (approved):
thanks for fixing this!

PR #297, opened 2023-06-02
Storage: Multipath support

benlwalker:
* Separate the path object from the controller, leaving open the possibility to have multiple paths.
* Discovery namespaces by connecting to controllers


storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -144,6 +201,42 @@ message NVMfRemoteControllerStatsResponse {
...
+    string parent = 1 [
+        (google.api.field_behavior) = REQUIRED,
+        (google.api.resource_reference).type = "opi_api.storage.v1/NVMfRemoteNamespace"
+    ];
+    int32 page_size = 2;
+    string page_token = 3;
+}
+
+message ListNVMfRemoteNamespacesResponse {
+    repeated NVMfRemoteNamespace namespaces = 1;
```
glimchb, line 216: should be `nv_mf_remote_namespaces`

glimchb (commented):
to fix linters you can apply this patch
```
diff --git a/storage/v1alpha1/backend_nvme_tcp.proto b/storage/v1alpha1/backend_nvme_tcp.proto
index 104da84..1f4b9b7 100755
--- a/storage/v1alpha1/backend_nvme_tcp.proto
+++ b/storage/v1alpha1/backend_nvme_tcp.proto
@@ -213,7 +213,7 @@ message ListNVMfRemoteNamespacesRequest {
 }

 message ListNVMfRemoteNamespacesResponse {
-    repeated NVMfRemoteNamespace namespaces = 1;
+    repeated NVMfRemoteNamespace nv_mf_remote_namespaces = 1;
     string next_page_token = 2;
 }
```

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -69,17 +92,51 @@ message NVMfRemoteController {
...
+    bool hdgst = 2;
+    bool ddgst = 3;
+    NvmeMultipath multipath = 4;
+    int64 io_queues_count = 5;
+    int64 queue_size = 6;
+}
+
+message NVMfPath {
+    // object's unique identifier
+    common.v1.ObjectKey id = 1;
```
glimchb, line 104: this should be:
```
    string name = 1;
```

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -69,17 +92,51 @@ message NVMfRemoteController {
...
+
+message NVMfRemoteNamespace {
+    // namespace's unique key. Assign to middle/front-end objects.
+    common.v1.ObjectKey id = 1;
+
+    // controller through which the namespace is visible
+    common.v1.ObjectKey controller_id = 2;
+
+    // NSID
+    int32 nsid = 4;
```
artek-koltun, line 127: int32 nsid = 4;
and subsequent

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -69,17 +92,51 @@ message NVMfRemoteController {
...
-    string subnqn = 6;
-    bool hdgst = 7;
-    bool ddgst = 8;
-    NvmeMultipath multipath = 9;
-    int64 io_queues_count = 10;
-    int64 queue_size = 11;
-    string hostnqn = 12;
+    bool hdgst = 2;
+    bool ddgst = 3;
+    NvmeMultipath multipath = 4;
```
jainvipin, line 97: Minor: we should call this field `NvmeMultipathType` instead of `NVMeMultipath`.

glimchb (changes requested):
please remove `storage/v1alpha1/google` folder from the commit

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -69,17 +92,62 @@ message NVMfRemoteController {
...
+    int64 io_queues_count = 5;
+    int64 queue_size = 6;
+}
+
+message NVMfPath {
+    // user can only set {resource}_id on the Create request object
```
glimchb, line 103: missing:
```
    option (google.api.resource) = {
        type: "storage.opiproject.org/NVMfPath"
        pattern: "volumes/{volume}"
    };
```

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -69,17 +92,62 @@ message NVMfRemoteController {
...
+    // Source port (e.g. Port of local NIC)
+    int64 source_trsvcid = 9;
+
+    // Host NQN
+    string hostnqn = 10;
+}
+
+message NVMfRemoteNamespace {
+    // namespace's unique key. Assign to middle/front-end objects.
+    common.v1.ObjectKey id = 1;
```
glimchb, line 132: id is incorrect, should be:
```
    option (google.api.resource) = {
        type: "storage.opiproject.org/NVMfRemoteNamespace"
        pattern: "volumes/{volume}"
    };
    
    string name = 1;
```

glimchb (changes requested):
this patch will fix CI failure:
```
diff --git a/storage/v1alpha1/backend_nvme_tcp.proto b/storage/v1alpha1/backend_nvme_tcp.proto
index 2c51925..413561b 100755
--- a/storage/v1alpha1/backend_nvme_tcp.proto
+++ b/storage/v1alpha1/backend_nvme_tcp.proto
@@ -67,10 +67,10 @@ service NVMfRemoteControllerService {

     rpc CreateNVMfPath (CreateNVMfPathRequest) returns (NVMfPath) {
         option (google.api.http) = {
-            post: "/v1/{parent=subsystems}"
+            post: "/v1/volumes"
             body: "nv_mf_path"
         };
-        option (google.api.method_signature) = "parent,nv_mf_path,nv_mf_path_id";
+        option (google.api.method_signature) = "nv_mf_path,nv_mf_path_id";
     }

     rpc DeleteNVMfPath (DeleteNVMfPathRequest) returns (google.protobuf.Empty) {
@@ -100,6 +100,11 @@ message NVMfRemoteController {
 }

 message NVMfPath {
+    option (google.api.resource) = {
+        type: "storage.opiproject.org/NVMfPath"
+        pattern: "volumes/{volume}"
+    };
+
     string name = 1;

@@ -128,8 +133,13 @@ message NVMfPath {
 }

 message NVMfRemoteNamespace {
-    // namespace's unique key. Assign to middle/front-end objects.
-    common.v1.ObjectKey id = 1;
+    option (google.api.resource) = {
+        type: "storage.opiproject.org/NVMfRemoteNamespace"
+        pattern: "volumes/{volume}"
+    };
+
+    // user can only set {resource}_id on the Create request object
+    string name = 1;

     // controller through which the namespace is visible
     common.v1.ObjectKey controller_id = 2;
@@ -229,10 +239,6 @@ message ListNVMfRemoteNamespacesResponse {
 }

 message CreateNVMfPathRequest {
-    string parent = 1 [
-        (google.api.field_behavior) = REQUIRED,
-        (google.api.resource_reference).type = "opi_api.storage.v1/NVMfPath"
-    ];
     NVMfPath nv_mf_path = 2 [(google.api.field_behavior) = REQUIRED];
     string nv_mf_path_id = 3;
 }
(END)
```

glimchb:
> In this case, each NVMfPath is a true NVMe-oF controller (a session) established between a host and a subsystem. These paths are aggregated into an object that behaves like a single NVMe-oF controller - that's what the NVMfController represents (the aggregated thing). An NVMfRemoteNamespace is a namespace - the thing you can read/write - not a controller.

@benlwalker looks like there are no RPCs to create/delete `NVMfRemoteNamespace`... so why have it as an object ? when it is created ? I guess when you connect to remote system with many namespaces, you probably want to create them internally, and then just use them when you reference that object in midllend/frontend... right ?

benlwalker:
> @benlwalker looks like there are no RPCs to create/delete `NVMfRemoteNamespace`... so why have it as an object ? when it is created ? I guess when you connect to remote system with many namespaces, you probably want to create them internally, and then just use them when you reference that object in midllend/frontend... right ?

If you imagine this is like the Linux kernel, there are no nvme-cli commands to create /dev/nvme0n1 (ignoring format namespace). Instead, you create /dev/nvme0 by connecting to a subsystem and it discovers the namespaces. You then list the namespaces and can mount filesystems on them. This is exactly the same thing. You discover the namespaces, then list them. 

They need to be individual objects because these are what the middle/front end reference when building a middle end bdev or frontend nvmf subsystem. A crypto middle end bdev, for example, would be built on top of one of these namespace objects. Again it's just like Linux (or SPDK) where you create an md raid or a dm-crypt volume on top of /dev/nvme0n1 (the namespace you discovered), not on top of /dev/nvme0 (the controller you created).

glimchb:
> They need to be individual objects because these are what the middle/front end reference when building a middle end bdev or frontend nvmf subsystem. A crypto middle end bdev, for example, would be built on top of one of these namespace objects. Again it's just like Linux (or SPDK) where you create an md raid or a dm-crypt volume on top of /dev/nvme0n1 (the namespace you discovered), not on top of /dev/nvme0 (the controller you created).

thanks, that is what I figured out as I was writing the question. perhaps similarity with `/dev/nvme0n1` and `/dev/nvme0` somewhere in the comments could help

artek-koltun (approved):
Looks ok for me. Let's try to implement it in SPDK bridge and see how it fits

artek-koltun (commented):
Worked on [initial implementation](https://github.com/opiproject/opi-spdk-bridge/pull/461) for this API change. I thought about a potential usage scenario: A CSI driver crashed and wants to find out all the paths existed before the crash. It would be nice to have List/Get calls for that purpose as well. However, we have only create and delete...

PR #308, opened 2023-07-06
[storage][backend] Add psk field to NvmeRemoteController

glimchb:
this will enable Nvme/TLS secure connection to the target


jainvipin (approved):
LGTM
While it can be argued that pre shared key for TLS need to be path specific, it doesn't makes sense to have different keys for different backend targets. And therefore keeping this in controller object looks good to me.

artek-koltun (approved):
looks ok for me

PR #310, opened 2023-07-10
subnet can have an associated tunnel interface for evpn

jainvipin:
@glimchb @venkatmahalingam - pls review one field change in subnet.proto file


glimchb (changes requested):
Vipin, I thought you adding subnet reference to the tunnel object. Not the other way around… if tunnel is vxlan, and subnet is bridge we may have one to many…

jainvipin:
> Vipin, I thought you adding subnet reference to the tunnel object. Not the other way around… if tunnel is vxlan, and subnet is bridge we may have one to many…

jainvipin:
@glimchb @sanders-mark  - should we merge this? this way cloud APIs can be a bit more flexible. Having said that we are addressing evpn use case as a first class APIs and so the use case for this enhancement becomes somewhat unclear.

PR #314, opened 2023-07-26
change id refs to name

jainvipin:
@glimchb - pls review; mostly the changes are mechanical. I will update the spdk refs and other pages using these protos subsequently.
Also missing is generated code for easier review, which I will add subsequently after LGTMs

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -7,6 +7,9 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  contents: read
+
 jobs:
   call:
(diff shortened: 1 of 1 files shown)


glimchb (commented):
looks good
there are still some `common.v1.ObjectKey` in the rest of protobufs...
we can address them now or later on...

jainvipin:
> looks good there are still some `common.v1.ObjectKey` in the rest of protobufs... we can address them now or later on...

I was hoping to address all nvme related things, because I don't have a way to test the others. If you are okay, we can do the rest later.

PR #315, opened 2023-07-26
feat(storage): change id refs to name

jainvipin:
refer #314 for the context


glimchb (commented):
lgtm

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -241,12 +240,12 @@ message GetNvmeRemoteControllerRequest {
 
 message NvmeRemoteControllerResetRequest {
     // object's unique identifier
```
artek-koltun, line 242: Is the comment still relevant? We have a full resource name incorporating resource id here. I'd propose a change to smth like `full resource name`.. WDYT?

glimchb (approved):
approved
will open an issue to add require attribute like this:
```
    string name = 1 [
        (google.api.field_behavior) = REQUIRED,
        (google.api.resource_reference).type = "opi_api.storage.v1/AioVolume"
    ];
```

storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -329,7 +328,7 @@ message GetNvmeSubsystemRequest {
 }
 
 message NvmeSubsystemStatsRequest {
-    common.v1.ObjectKey subsystem_id = 1;
```
glimchb, line 332: @jainvipin I missed this. I should be simple `name` field

PR #317, opened 2023-08-01
Storage API changes with name and field_behavior

glimchb:
- feat(storage): fill field_behavior for resource name
- feat(storage): fix field numbers
- feat(storage): use name key in statistics
- feat(storage): leftover fixes to fields and numbers
- feat(storage): change from OPTIONAL to OUTPUT_ONLY
- feat(storage): fix AIP core::0127:resource-name-extraction
- feat(storage): fix AIP core::0127::http-annotation
- feat(storage): auto-generate protos
- refactor(storage): add IMMUTABLE behavior for resource name field 

see https://google.aip.dev/203

### DIFF
--- a/storage/Makefile
+++ b/storage/Makefile
@@ -16,5 +16,5 @@ all:
 	mv google "${PWD}"/v1alpha1/
 	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh -v "${PWD}/../common/v1":/common -v "${PWD}"/v1alpha1/:/out -w /out -v "${PWD}"/v1alpha1:/protos pseudomuto/protoc-gen-doc -c "protoc -I /common -I /protos --doc_out=/out --doc_opt=markdown,autogen.md /protos/*.proto /common/*.proto"
 	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh -v "${PWD}/../common/v1":/common -v "${PWD}"/v1alpha1/:/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.36.0 -c "api-linter -I /common /out/*.proto --output-format summary"
-	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh -v "${PWD}/../common/v1":/common -v "${PWD}"/v1alpha1/:/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.36.0 -c "api-linter -I /common /out/*.proto --output-format github --disable-rule=core::0192::has-comments --disable-rule=core::0156 --set-exit-status"
+	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh -v "${PWD}/../common/v1":/common -v "${PWD}"/v1alpha1/:/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.36.0 -c "api-linter -I /common /out/*.proto --output-format github --disable-rule=core::0192::has-comments --set-exit-status"
 	rm -rf "${PWD}"/v1alpha1/google


storage/v1alpha1/backend_aio.proto
```diff
@@ -67,7 +71,10 @@ message AioVolume {
     // name is an opaque object handle that is not user settable.
     // name will be returned with created object
     // user can only set {resource}_id on the Create request object
-    string name = 1;
+    string name = 1 [
+        (google.api.field_behavior) = OUTPUT_ONLY,
```
artek-koltun, line 75: Is `OUTPUT_ONLY` a correct annotation here?
This is a hard one... According to the annotation description:
```
including the field in a message in a request does nothing (the server must clear out any value in this field and must not throw an error as a result of the presence of a value in this field on input). Similarly, services must ignore the presence of output only fields in update field masks.
```
Based on that, a user cannot specify a `name`, since the server must ignore it. But this resource is used in Update call and `name` is used to find that resource internally...

This might be a resolution: 2 weeks ago [a PR](https://github.com/aip-dev/google.aip.dev/pull/1180) from aip team member was created where it is said that `name` field must not be annotated.

intelfisz:
@glimchb - please change the commit msgs for fixes to have fix(storage) as prefix

artek-koltun:
> @glimchb - please change the commit msgs for fixes to have fix(storage) as prefix

There are some guidance from angular [here](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) based on that:
feat(storage): fill field_behavior for resource name -> `refactor/feat`
feat(storage): fix field numbers -> `fix`
feat(storage): use name key in statistics -> `refactor/feat`
feat(storage): leftover fixes to fields and numbers -> `refactor/feat`
feat(storage): change from OPTIONAL to OUTPUT_ONLY -> `refactor/feat`
feat(storage): fix AIP core::0127:resource-name-extraction -> `ci`
feat(storage): fix AIP core::0127::http-annotation -> `ci`
feat(storage): auto-generate protos -> `chore`? it is a special category in Conventional commit we use for such housekeeping tasks

storage/v1alpha1/backend_aio.proto
glimchb: yep, I saw many discussion around `name`
I think let's start more strict with `OUTPUT_ONLY` and will re-consider in the future. Opened #318 to track it

artek-koltun: If we start stricter and wait till that PR is officially merged, then let's add `IMMUTABLE` to `OUTPUT_ONLY` like Google do for names [here](https://github.com/googleapis/googleapis/blob/449686b58eaee7df382c2020f5069769deebde17/google/cloud/dataqna/v1alpha/question.proto#L43-L48) and [here](https://github.com/googleapis/googleapis/blob/449686b58eaee7df382c2020f5069769deebde17/google/cloud/tpu/v1/cloud_tpu.proto#L261-L265)

glimchb:
> @glimchb - please change the commit msgs for fixes to have fix(storage) as prefix

done, thanks for suggestion

storage/v1alpha1/backend_aio.proto
glimchb: I need to read more about `IMMUTABLE` before adding it....

artek-koltun: Current AIP [says](https://google.aip.dev/203#immutable) about `IMMUTABLE`:
```
Potential use cases for immutable fields (this is not an exhaustive list) are:

    Names or IDs which are set on creation and then used as a primary key.
```

artek-koltun (approved):
Looks good at the moment

PR #321, opened 2023-08-08
Set annotations for storage frontend according to AIP

artek-koltun:
Used AIPs: [203](https://google.aip.dev/203#input-only), [149](https://google.aip.dev/149)

Some considerations applied in this patch:
1. `nvme_controller_id` is a zero-based value. Since a default value for `int` field is zero, it is hard to distinguish if a client requested to set a controller id to 0, or a value was not specified at all and was set by default to 0. For such a case, `optional` label was used as outlined in [AIP-149](https://google.aip.dev/149).

2. `PciEndpoint` was set as `REQUIRED` to avoid ambiguity. That field is used by all the bridges to create different devices.
* Later, if we decide that bridges must be able to allocate values automatically to a `PciEndpoint` we can mark it as `OPTIONAL` and this action is treated as [backward compatible](https://google.aip.dev/203#backwards-compatibility)
* We use zero-based values for `PciEndpoint` fields like `virtual_function`, based on [AIP-149](https://google.aip.dev/149), we should  mark those fields with `optional` label. However, it leads to a linter error complaining on optional label and `REQUIRED` annotation, despite the fact AIP-149 explicitly [allows](https://google.aip.dev/149#field-behavior-and-optional) such a case. As a workaround, all field types within `PciEndpoint` are changed to `google.protobuf.Int32Value`.


storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -205,16 +205,16 @@ message NvmeController {
...
     // status holds server generated values
-    NvmeControllerStatus status = 3;
+    NvmeControllerStatus status = 3 [(google.api.field_behavior) = OUTPUT_ONLY];
 }
 
 message NvmeControllerSpec {
     // subsystem controller id range: 0 to 65535.
     // must not be reused under the same subsystem
-    int32 nvme_controller_id  = 1;
+    optional int32 nvme_controller_id  = 1 [(google.api.field_behavior) = OPTIONAL];
```
glimchb, line 217: why `optional int32`  prefix and not just annotation ?

glimchb (commented):
good work, thanks! will let others review and comment as well

storage/v1alpha1/frontend_nvme_pcie.proto
artek-koltun: I described it in the PR description, but in general: `optional` label is a way to distinguish if a value was set intentionally to zero by a client or not. `nvme_controller_id` is such a case since it is a zero-based value
See [AIP-149](https://google.aip.dev/149)

sburla-marvell (approved):

intelfisz (approved):

glimchb:
waiting on @jainvipin before merge

storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -159,34 +159,34 @@ message NvmeSubsystem {
...
 
     // maximum namespaces within a subsystem
-    int64 max_namespaces = 4;
+    int64 max_namespaces = 4 [(google.api.field_behavior) = OPTIONAL];
 }
 
 message NvmeSubsystemStatus {
     // firmware revision, must not exceed 'NSV_CTRLR_FW_REV_LEN'
-    string firmware_revision  = 1;
+    string firmware_revision  = 1 [(google.api.field_behavior) = OUTPUT_ONLY];
```
jainvipin, line 186: if `NvmeSubsystemStatus` is `OUTPUT_ONLY`, do we need redundancy in all the fields to be `OUTPUT_ONLY`?

glimchb: @artek-koltun please comment

artek-koltun: It is[ recommended](https://google.aip.dev/203#output-only) to annotate:
```
Additionally, a field should only be described as output only if it is a field
in a resource message, or a field of a message farther down the tree.
```
`NvmeSubsystemStatus`'s fields are farther down the tree of a resource message `NvmeSubsystem`

PR #334, opened 2023-08-25
Add presentation for Netlink and YNL

donaldh:


sandersms:
@donaldh Can you update your checking to have docs: at the beginning of the commit.  The checks on the commits now have the addition of checking for the type as part of the checks.

PR #342, opened 2023-08-29
feat(evpn): annotate field_behavior per AIP 203 and 123

glimchb:
- feat(evpn): annotate field_behavior per AIP 203
- feat(evpn): annotate singular and plural per AIP 123
- chore(evpn): auto-generated protos


glimchb:
Those are must changes to pass new AIP linter, see #341

PR #343, opened 2023-08-30
Feat backend separate transport specific messages

artek-koltun:
For continuation of the discussion in https://github.com/opiproject/opi-api/pull/330


storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -170,25 +175,31 @@ message NvmePath {
...
 
     // Destination address (e.g. IP address, BDF for local PCIe)
-    string traddr = 5 [(google.api.field_behavior) = REQUIRED];
+    string traddr = 4 [(google.api.field_behavior) = REQUIRED];
 
+    // Not applicable for local PCIe. Required for Nvme over fabrics transport types
+    FabricsPath fabrics = 5 [(google.api.field_behavior) = OPTIONAL];
+}
+
+message FabricsPath {
```
glimchb, line 186: seems like we have tuple: (addr, port, nqn) for target and again (addr, port, nqn) for source, no ?

maybe we can somehow capture this in it's own message ?

benlwalker: The traddr for the target is in the common part (it's the BDF for PCIe), so it needs to be split across two objects here.

benlwalker (approved):

storage/v1alpha1/backend_nvme_tcp.proto
glimchb: @benlwalker WDYT about `struct spdk_nvme_transport_id` ?
https://github.com/spdk/spdk/blob/ee9fb405bfa6f11b75788c38752b2494598bbbd8/include/spdk/nvme.h#L455-L501

I think it is pretty good abstraction what we want to adopt here, why not ?

benlwalker: I do like that abstraction. To make that work here you'd need to have a FabricsPath with a source and target transport id, and then a MemoryBusPath with just a target transport id (or just traddr), and put the two into a OneOf.

glimchb (approved):
lgtm

PR #345, opened 2023-08-30
feat(storage): annotate singular and plural per AIP 123

glimchb:
see https://google.aip.dev/123#singular-and-plural
in preparation for new linter #341


sandersms (approved):
Changes LGTM

artek-koltun (approved):
LGTM

PR #352, opened 2023-09-07
docs(evpn): add sections to readme to have same look and feel

glimchb:


mardim91 (changes requested):
We do not have an implementation for pydpu only for godpu.

glimchb:
> We do not have an implementation for pydpu only for godpu.

not exactly accurate.
indeed there is no implementation for pydpu
however, we do package and distribute protobufs with https://pypi.org/project/pydpu/
since python is different from golang and need to have package with generated protobufs

 we also have examples if somebody wants to pick up and controbute full code for pydpu
https://github.com/opiproject/pydpu/blob/main/pydpu/evpn.py

that's why I think it still OK and useful to mention pydpu on this page

sandersms (approved):
Looks like a good start toward a general template view

PR #375, opened 2023-09-27
feat(storage): frontend traget case API

artek-koltun:


storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -240,34 +240,45 @@ message NvmeControllerSpec {
...
 
-    // xPU's PCI ID for the controller
-    PciEndpoint pcie_id = 3 [(google.api.field_behavior) = REQUIRED];
+    // transport type
+    NvmeTransportType trtype = 2 [(google.api.field_behavior) = REQUIRED];
+
+    // xPU's PCI/Fabrics endpoint for the controller.
+    // One of the values is required
+    // see https://github.com/aip-dev/google.aip.dev/issues/1147 for field_behavior annotations
+    oneof endpoint {
```
artek-koltun, line 249: A question: would it make sense to use a similar oneof structure in backend nvme for symmetry?

storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -240,34 +240,45 @@ message NvmeControllerSpec {
...
+        // Required for pcie transport type.
+        PciEndpoint pcie_id = 3 [(google.api.field_behavior) = OPTIONAL];
+
+        // Required for Nvme over fabrics transport types
+        FabricsEndpoint fabrics_id = 4 [(google.api.field_behavior) = OPTIONAL];
```
sburla-marvell, line 254: I might be missing something, I assume this structure is used in CreateNVMeController. The PCIendpoint struct in case of frontend is used to indicate the port, pf, vf of the NVMe controller that is being exposed by xPU over PCIe bus to a PCIe host.  In case of Fabrics endpoint, if this is run on xPU, then it looks like it will have to setup a NVMecontroller for fabric storage access for the xPU and not the host.  right ?

storage/v1alpha1/frontend_nvme_pcie.proto
glimchb: Yes, instead of optional fabrics field

storage/v1alpha1/frontend_nvme_pcie.proto
artek-koltun: Right. By this new fabrics endpoint, we setup a new NVMe-oF controller to expose for example local SSDs over a network(JBOF scenario). Thus, we can connect to that NVMe-oF controller from another xPU by using backend API

storage/v1alpha1/frontend_nvme_pcie.proto
artek-koltun: Ok, I will update backend in a separate PR then

storage/v1alpha1/frontend_nvme_pcie.proto
sburla-marvell: Ok. thank you. can that be added as comment here ?

sburla-marvell (approved):

glimchb (approved):
lgtm
looking forward for BackEnd change following this one to use all the common objects now

storage/v1alpha1/frontend_nvme_pcie.proto
artek-koltun: Updated the comment

storage/v1alpha1/opicommon.proto
```diff
@@ -43,6 +43,18 @@ message PciEndpoint {
     google.protobuf.Int32Value  virtual_function = 3 [(google.api.field_behavior) = REQUIRED];
 }
 
+// Represents Fabrics Endpoint
+message FabricsEndpoint {
```
glimchb, line 47: nvme: consider combining `FabricsEndpoint` and `FabricsPath`

since we want to extend this message with more fields anyways

Issue #376, opened 2023-09-29
nvme: consider combining `FabricsEndpoint` and `FabricsPath`


glimchb:
nvme: consider combining `FabricsEndpoint` and `FabricsPath`

since we want to extend this message with more fields anyways

_Originally posted by @glimchb in https://github.com/opiproject/opi-api/pull/375#discussion_r1341584567_

Issue #363, opened 2023-09-19
Make storage backend nvme file and content consistent with provided services


artek-koltun:
At the moment, we have an empty [backend_nvme_pcie.proto](https://github.com/opiproject/opi-api/blob/main/storage/v1alpha1) and [backend_nvme_tcp.proto](https://github.com/opiproject/opi-api/blob/main/storage/v1alpha1/backend_nvme_tcp.proto). However, it is planned to have fabrics and pcie in one backend service. Probably, rename of `backend_nvme_tcp.proto` to `backend_nvme.proto` and deletion of `backend_nvme_pcie.proto` makes sense.

Also, since we can connect to local pcie nvme drives by the service in `backend_nvme_tcp.proto` it might make sense to rename messages:
`NvmeRemoteController*`  -> `NvmeBackendController*`
`NvmeRemoteNamespace*`  -> `NvmeBackendNamespace*`

Additionally, see https://github.com/opiproject/opi-api/pull/343

artek-koltun:
Files are renamed/deleted

Message renames are left

PR #380, opened 2023-10-04
refactor: combine frontend and backend messages

artek-koltun:
resolves https://github.com/opiproject/opi-api/issues/376


storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -213,32 +213,32 @@ message NvmePath {
...
 // Represents Fabrics transport path parameters
 message FabricsPath {
-    // Destination service id (e.g. Port)
-    int64 trsvcid = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // Destination fabrics endpoint
+    FabricsEndpoint dst = 1 [(google.api.field_behavior) = REQUIRED];
 
     // Subsystem NQN
     string subnqn = 2 [(google.api.field_behavior) = REQUIRED];
```
artek-koltun, line 235: Want to discuss a possibility to move `subnqn` from `NvmePath` into `NvmeRemoteController`. I see `NvmeRemoteController` as a view into a remote subsystem, which can be reached by means of multiple `NvmePath`s. However, 2 `NvmePath`s  from a single `NvmeRemoteController` cannot lead to different subsystems, so, in my opinion, it would make sense to put `subnqn` into the controller to state a subsystem we want to reach by means of multiple paths.

The same about `hostnqn`. We can state our `hostnqn` when we define our view to the subsystem.

@benlwalker could you please comment?

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -213,32 +213,32 @@ message NvmePath {
...
-    // Address family
-    NvmeAddressFamily adrfam = 3 [(google.api.field_behavior) = REQUIRED];
-
-    // Source address (e.g. IP of local NIC)
-    string source_traddr = 4 [(google.api.field_behavior) = OPTIONAL];
-
-    // Source port (e.g. Port of local NIC)
-    int64 source_trsvcid = 5 [(google.api.field_behavior) = OPTIONAL];
+    // Source fabrics endpoint
+    FabricsEndpoint src = 3 [(google.api.field_behavior) = OPTIONAL];
```
artek-koltun, line 238: `source_traddr` and `source_trsvcid` are incorporated into this field. However, [FabricsEndpoint](https://github.com/opiproject/opi-api/blob/fc749d07fdda571db3a4747e56715657da920fd9/storage/v1alpha1/opicommon.proto#L55) contains required `adrfam` field. For example, in SPDK, it won't be used. There we just need to pass `hostaddr` and `hostsvcid`. But we can use this `adrfam` to pre-validate those values within bridges. As an alternative, we can make `adrfam` optional

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -213,32 +213,32 @@ message NvmePath {
...
-    string traddr = 4 [(google.api.field_behavior) = REQUIRED];
-
-    // Not applicable for local PCIe. Required for Nvme over fabrics transport types
-    FabricsPath fabrics = 5 [(google.api.field_behavior) = OPTIONAL];
+    // xPU's PCI/Fabrics endpoint for the controller.
+    // One of the values is required
+    // see https://github.com/aip-dev/google.aip.dev/issues/1147 for field_behavior annotations
+    oneof path {
+        // Required for pcie transport type. Shall contain BDF
+        string pcie = 4 [(google.api.field_behavior) = OPTIONAL];
```
artek-koltun, line 221: Should we try to use [PciEndpoint](https://github.com/opiproject/opi-api/blob/fc749d07fdda571db3a4747e56715657da920fd9/storage/v1alpha1/opicommon.proto#L26) here for symmetry with [frontend](https://github.com/opiproject/opi-api/blob/fc749d07fdda571db3a4747e56715657da920fd9/storage/v1alpha1/frontend_nvme_pcie.proto#L249)?

glimchb: Yes, absolutely

glimchb (changes requested):
Comment inside in pcie end point 
And I think we can finally remove 
`backend_nvme_pcie.proto`
and rename the `backend_nvme_tcp.proto` to `backend_nvme.proto`

storage/v1alpha1/backend_nvme_tcp.proto
```diff
@@ -213,32 +213,32 @@ message NvmePath {
...
+
+        // Required for Nvme over fabrics transport types
+        FabricsPath fabrics = 5 [(google.api.field_behavior) = OPTIONAL];
+    }
 }
 
 // Represents Fabrics transport path parameters
 message FabricsPath {
```
glimchb, line 229: I was thinking we get rid of `FabricsPath` and use only `FabricsEndpoint` it is too confusing to have them both...

artek-koltun:
>And I think we can finally remove `backend_nvme_pcie.proto` and rename the `backend_nvme_tcp.proto` to `backend_nvme.proto`

Previously, https://github.com/opiproject/opi-api/issues/363 was created to address that change. Covered in https://github.com/opiproject/opi-api/pull/382

glimchb:
merged #382

storage/v1alpha1/backend_nvme_tcp.proto
artek-koltun: Done. Added comments how `PcieEndpoint` values correspond to BDF and added `domain_id`

storage/v1alpha1/backend_nvme_tcp.proto
artek-koltun: Removed. Now the field behavior annotations are a bit less restrictive, but it may be ok for us

artek-koltun: `host*` members are moved to `FabricsEndpoint`, but would it make sense to move it to `NvmeRemoteControlelr`, since they are only backend related parameters?

PR #381, opened 2023-10-04
feat(storage): add hostnqn ans psk for frontend subsystem

artek-koltun:


storage/v1alpha1/frontend_nvme_pcie.proto
```diff
@@ -198,6 +198,12 @@ message NvmeSubsystemSpec {
 
     // maximum namespaces within a subsystem
     int64 max_namespaces = 4 [(google.api.field_behavior) = OPTIONAL];
+
+    // host NQN
+    string hostnqn = 5 [(google.api.field_behavior) = OPTIONAL];
```
artek-koltun, line 203: we can start with a stricter rule allowing only one host. Later, if needed, we can allow multiple hosts by using
```
// host NQNs with corresponding PSK keys. Empy string value if encryption is not used
map<string, bytes> hostnqns = 5[(google.api.field_behavior) = OPTIONAL];
```

glimchb (approved):

PR #382, opened 2023-10-05
refactor(storage): nvme proto file restructure

artek-koltun:
see https://github.com/opiproject/opi-api/issues/363


glimchb (approved):
great, approved

PR #383, opened 2023-10-05
refactor(storage): frontend cleanup pci leftovers

artek-koltun:
`FrontendNvmeService` serves pcie and fabrics devices. Remove pci from names and comments.

got https://linter.aip.dev/216/nesting lint error and had to move state enum into namespace status message


glimchb (approved):

glimchb:
is https://linter.aip.dev/216/nesting new ? how come we were not failing it before ?

artek-koltun:
> is https://linter.aip.dev/216/nesting new ? how come we were not failing it before ?

The name of the field changed to `state` and triggered the check. This check was run against only `State state = 1` However, I decided to move `OperState` enum for consistency as well since it is used only in that structure

PR #395, opened 2023-10-24
fix(buf): fix buf linting for common, inventory, security

sandersms:
Corrects lint errors reported by buf for the common, inventory, and security folders that are enabled.

Additional folders can be added as the buf reported issues are resolved for those areas in turn.


glimchb (approved):
great work!

PR #413, opened 2023-12-06
refactor(cloud): update to support buf lint and generation automation

sandersms:
Transition cloud to the buf lint and buf generate operation to allow for automation of the protobuf file generation.

Support for cpp and java are being removed at this time. Generation supports go and python. Support for other languages can be added as needed by the buf.gen.yaml file.

Be aware that the .proto files are moved up from the v1alpha1 directory to support the buf operation and to keep the version of the generated files in a separate (v1, v1alpha1, etc.) directory structure. This will allow for the future ability to autogenerate the files and place them in a separate repo location.


network/cloud/Makefile
```diff
@@ -2,22 +2,23 @@
 # Copyright (C) 2022 Intel Corporation
 # Copyright (c) 2022 Dell Inc, or its subsidiaries.
 
-all:
+all: buflint doc
```
artek-koltun, line 5: Please see my comments to 
https://github.com/opiproject/opi-api/pull/411

network/cloud/cloudrpc.proto
```diff
@@ -879,10 +889,15 @@ message DeleteVnicRequest {
 
 // Update vnic request
 message UpdateVnicRequest {
+    // Name of Vnic requested
+    string name = 1 [
```
artek-koltun, line 893: To be honest, these `Update*` changes look strange...
The resource (`vnic`) already contains name and adding it one more time is suspicious. I assume we can resolve it on the annotation level. Can you try to annotate `UpdateVnic` the following way
```proto
rpc UpdateVnic (UpdateVnicRequest) returns (Vnic) {
        option (google.api.http) = {
            patch: "/v1/{vnic.name=vnics/*}"
            body: "vnic"
        };
        option (google.api.method_signature) = "vnic, update_mask";
    }
```

sandersms: Yes, much cleaner solution using the different annotate.

network/cloud/bgp.proto
```diff
@@ -354,7 +354,7 @@ message BGPNLRIPrefixStatus {
   int32 flap_starttime = 19;
   // If bgpNlriPrefixBest is 'true', then this field is set to 'routeIsBest'.
   // Otherwise, it reports the stage in the decision process when the route was determined to be non-best.
-  BGPRsnNotBest reason_not_best = 20;
+  BGPRouteReason reason_not_best = 20;
```
artek-koltun, line 357: we could consider renaming the member as well

sandersms: > we could consider renaming the member as well

We could change it to route_reason = 20; as an option.  The intent is to report the reason for not having the best route which really is the reason for the particular route selection where one of the reason's is "Route is best".  I will let @jainvipin respond on the changing since the elements are from his contribution and his input on the enum and parameter should be considered on this one.

network/cloud/cloudrpc.proto
```diff
@@ -1604,7 +1604,6 @@ message DeleteSubnetRequest {
 
 // Update subnet request
 message UpdateSubnetRequest {
-    // updated subnet info
```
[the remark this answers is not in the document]
sandersms: Enabled buf COMMENTS linting rule in latest update with two comment checks disabled currently to be worked in later updates.

artek-koltun (approved):
LGTM!

network/cloud/bgp.proto
jainvipin, line 357: agree with @artek-koltun - we should change the field/member name as well to reflect its intent; thanks for fixing this.

jainvipin (commented):
@sandersms - all good; few minor comments inline... will approve after those fixes.

network/cloud/bgp.proto
```diff
@@ -106,7 +106,7 @@ message BGPPeerSpec {
   // send extended community attributes to neighbor
   bool send_ext_comm = 6;
   // peer is a route reflector client
-  BGPPeerRRClient rr_client = 7;
+  BGPPeerRR rr_client = 7;
```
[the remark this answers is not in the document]
sandersms: I will add the Client back to the enum and the parameter.  It was the linting checking on the enum definitions that resulted in the change.

Issue #332, opened 2023-08-23
code generator protoc-gen-grpc-gateway hasn't been updated to support optional fields in proto3


chujieyang:
see https://github.com/grpc-ecosystem/grpc-gateway/issues/1278
see https://github.com/namely/docker-protoc/pull/361

An exception occurs when I use namely/protoc-all:1.51_2 with the parameter: --with-gateway to convert a proto file to a REST service:

```
... is a proto3 file that contains optional fields, but code generator protoc-gen-grpc-gateway hasn't been updated to support optional fields in proto3. Please ask the owner of this code generator to support proto3 optional.--grpc-gateway_out:
make: *** [Makefile:9: all] Error 1

```
Would like to update the protoc version in the namely/protoc-all mirror to support this.

glimchb:
trying to enable grpc gw for inventory #359

PR #419, opened 2023-12-15
refactor(storage): update to support buf lint and generation automation

artek-koltun:
Transition storage to the buf lint and buf generate operation to allow for automation of the protobuf file generation.

Support for cpp and java are being removed at this time. Generation supports go and python. Support for other languages can be added as needed by the buf.gen.yaml file.

Be aware that the .proto files are moved up from the v1alpha1 directory to support the buf operation and to keep the version of the generated files in a separate (v1, v1alpha1, etc.) directory structure. This will allow for the future ability to autogenerate the files and place them in a separate repo location.

uuid message is replaced to string type, based on the internal discussion and an intention to remove common package at all.

grpc-gateway is generated as a part of the build, so fixes https://github.com/opiproject/opi-api/issues/332


sandersms (approved):
Changes look good - same as the ones I was working with the addition of the removal of the uuid.proto and the common directory.

Issue #86, opened 2022-08-09
Network: Expand use case(s) for networking


sandersms:
1) Define the requirements for each use case
2) Address the usage of the OVS DB, Openflow, OpenConfig, P4, etc.
3) Map the use case with the various APIs that need to be considered to support the use of OVS, SONiC, VPP, SONiC in the networking functionality that can reside in the xPU
4) Include operations such as switching, routing, firewall in the use cases and API
5) Include the scope for ports, links, vnets, etc.

jainvipin:
cloud API proposal in #128

PR #443, opened 2024-02-23
docs(network): update readme to mermaid diagrams

sandersms:
Add additional use case content
Cleanup diagrams using drawio direct file inclusion

https://github.com/opiproject/opi-api/issues/86


glimchb (changes requested):
`Diagram error not found` error on the readme file when you click view file

sandersms:
> `Diagram error not found` error on the readme file when you click view file see https://github.com/opiproject/opi-api/blob/e6de02f10ddb5e1479f53c4584c8c6f08556b055/network/README.md#openconfig-model

Thanks, I must have missed removing that as a residual.  Will get that fixed now.

glimchb (approved):

PR #445, opened 2024-02-26
feat: (EVPN GW API) Add "components" field to each API object's status

mardim91:
This is an API enhancement for the EVPN GW use case.

The addition of a "components" field in the status sector of each API object will enable the capture of the state and details that are generated by the different subscribed components (FRR component, Linux component, etc...) which are realizing the configuration for each API object.

This addition follows the re-architecture of the opi-evpn-br which has been described [here](https://github.com/opiproject/opi-evpn-bridge/blob/main/docs/OPI-EVPN-BRIDGE-Arch.pptx)


sandersms (approved):
Component addition looks fine for the tracking of status

mardim91:
Thank you @sandersms

Issue #446, opened 2024-03-08
[security] Update IPSec protobuf API to CRUD model


sandersms:
The current IPSec protobuf API follows a service model of specific rpc commands instead of a CRUD model.  The API needs to be refactored to the Create, Update, List, Delete, Get structure of commands.  This will allow for various offloads of IPSec to be supported.

The current support is focused toward strongswan and other options for IPSec support need to be accommodated.  Other options for IPSec could be openswan, racoon, etc.

sandersms:
We may want to expand this for VPNs in general or create an API that addresses VPNs to cover alternatives such as wireguard, OpenVPN, and others that are alternatives to IPSec and IKEv2.

Issue #104, opened 2022-08-29
security: please separate reference implementation VS generic APIs


glimchb:
looking at this section of the readme file : https://github.com/opiproject/opi-api/blob/main/security/README.md#architecture

I do think we need to separate reference implementation VS generic APIs...
`vici` on this page is just an example... 
everybody else are free to use their own implementation instead of strongswan or vici...
see the picture from https://github.com/opiproject/opi-poc/blob/main/storage/OPI-storage-SPDK-bridge.png

also storage in this repo doesn't talk about specific implementations [storage readme]([../storage/README.md)

_Originally posted by @glimchb in https://github.com/opiproject/opi-api/issues/102#issuecomment-1230464554_

glimchb:
fyi @shorman-corigine

sandersms:
Close issue with CRUD model addition on July 1, 2024 checkin.

PR #453, opened 2024-06-24
Change security API to a CRUD API

stevedoyle:
Resolves issue #446 and #104

Update the OPI security API to use CRUD operations and to use more non-implementation specific types. The v1 API messages were very specifc to a strongSwan implementation. This version updates the API to align with the more generic types from the IETF yang model described in RFC 9061.

Using an API version tag of v2alpha1 as the API is not backwards compatible with the v1 IPsec APIs.

Adpoting protobuf naming conventions for all messages and services.

BREAKING-CHANGE: Breaks compatiblity with the existing OPI security API definition.


artek-koltun:
@stevedoyle pls see failed CI jobs:
https://github.com/opiproject/opi-api/actions/runs/9646431855/job/26602822438?pr=453
https://github.com/opiproject/opi-api/actions/runs/9646431854/job/26602822702?pr=453

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+  rpc ListIkePeers(ListIkePeersRequest) returns (ListIkePeersResponse) {}
+
+  // Create an IKE connection. The request includes specification of the local
+  // and remote IKE peers and the specification of the IPsec SAs (aka child SAs)
+  // from this IKE connection.
+  rpc CreateIkeConn(CreateIkeConnRequest) returns (CreateIkeConnResponse) {}
+  // Update an existing IKE connection.
+  rpc UpdateIkeConn(UpdateIkeConnRequest) returns (UpdateIkeConnResponse) {}
+  // Delete an existing IKE connection.
+  rpc DeleteIkeConn(DeleteIkeConnRequest) returns (DeleteIkeConnResponse) {}
```
sandersms, line 45: Delete action can return "google.protobuf.Empty" instead.  Then remove the DeleteIkeConnResponse message.

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+  rpc GetIkeConn(GetIkeConnRequest) returns (GetIkeConnResponse) {}
+  // List existing IKE connections
+  rpc ListIkeConns(ListIkeConnsRequest) returns (ListIkeConnsResponse) {}
+
+  // Create an IPsec Security Association
+  rpc CreateIpsecSa(CreateIpsecSaRequest) returns (CreateIpsecSaResponse) {}
+  // Update an existing IPsec Security Association
+  rpc UpdateIpsecSa(UpdateIpsecSaRequest) returns (UpdateIpsecSaResponse) {}
+  // Delete an existing IPsec Security Association
+  rpc DeleteIpsecSa(DeleteIpsecSaRequest) returns (DeleteIpsecSaResponse) {}
```
sandersms, line 56: Delete action can return "google.protobuf.Empty" instead.  Then remove the DeleteIpsecSaResponse message.

security/Makefile
```diff
@@ -2,22 +2,29 @@
 # Copyright (C) 2022 Intel Corporation
 # Copyright (c) 2022 Dell Inc, or its subsidiaries.
 
+APIVER = v2alpha1
```
artek-koltun, line 5: do we already have a policy how we manage versions? I no, I'd keep it as it was `v1alpha1`
@sandersms WDYT?

security/Makefile
```diff
@@ -2,22 +2,29 @@
...
-	rm -f ./v1/autogen.md
-	mkdir -p  ./v1
+	rm -f ./${APIVER}/autogen.md
+	mkdir -p  ./${APIVER}
 
 	# protoc doesn't include annotation and http googleapis, so we have to get them here
 	curl -kL https://github.com/googleapis/googleapis/archive/master.tar.gz | tar --strip=1 -zxvf - googleapis-master/google/api
 
-	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh -v "${PWD}"/v1/:/out -w /out -v "${PWD}":/protos pseudomuto/protoc-gen-doc:1.5.1 -c "protoc -I /protos --doc_out=/out --doc_opt=markdown,autogen.md /protos/*.proto"
+	docker run --user=$$(id -u):$$(id -g) --rm --entrypoint=sh \
```
artek-koltun, line 17: Can I ask you to update Makefiles in other folders, so they look and feel the same?

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+  // Update an existing IKE peer specification.
+  rpc UpdateIkePeer(UpdateIkePeerRequest) returns (UpdateIkePeerResponse) {}
+  // Delete an existing IKE peer specification.
+  rpc DeleteIkePeer(DeleteIkePeerRequest) returns (DeleteIkePeerResponse) {}
+  // Get an existing IKE peer specification.
+  rpc GetIkePeer(GetIkePeerRequest) returns (GetIkePeerResponse) {}
+  // List existing IKE peers.
+  rpc ListIkePeers(ListIkePeersRequest) returns (ListIkePeersResponse) {}
+
+  // Create an IKE connection. The request includes specification of the local
```
artek-koltun, line 38: For other domains, we have one service per a resource.
Could you please create dedicated services per IKE Peer, IKE Connection, IPSec SA

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+service IpsecService {
+  // Create an IKE peer. This request includes the specification of the keys and certificates
+  // associated with the peer.
+  rpc CreateIkePeer(CreateIkePeerRequest) returns (CreateIkePeerResponse) {}
+  // Update an existing IKE peer specification.
+  rpc UpdateIkePeer(UpdateIkePeerRequest) returns (UpdateIkePeerResponse) {}
+  // Delete an existing IKE peer specification.
+  rpc DeleteIkePeer(DeleteIkePeerRequest) returns (DeleteIkePeerResponse) {}
+  // Get an existing IKE peer specification.
+  rpc GetIkePeer(GetIkePeerRequest) returns (GetIkePeerResponse) {}
```
artek-koltun, line 34: Gets should return [the corresponding resources](https://google.aip.dev/131). Here we need to return `IkePeer`
Pls also update the rest

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+option go_package = "github.com/opiproject/opi-api/security/v2alpha1/gen/go";
+
+// The IPsec service defines operations on:
+// - IKE Peer Association Database (PAD)
+// - IKE Connections
+// - IPsec Security Associations (SAs)
+service IpsecService {
+  // Create an IKE peer. This request includes the specification of the keys and certificates
+  // associated with the peer.
+  rpc CreateIkePeer(CreateIkePeerRequest) returns (CreateIkePeerResponse) {}
```
artek-koltun, line 28: We are missing annotations for methods and messages. For Create, pls see: https://google.aip.dev/133

You can look at network/storage to see what annotations are needed

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+}
+
+// IKEv2 Fragmentation, as per RFC 7383. If IKEv2 fragmentation is enabled, it
+// is possible to specify the MTU.
+message IkeFragmentation {
+  // Enable/Disable IKEv2 fragmentation.
+  bool enabled = 1;
+  // When fragmentation is enabled, the MTU that IKEv2 can use for IKEv2
+  // fragmentation.
+  optional uint32 mtu = 2;
```
artek-koltun, line 215: use of signed types is discouraged: https://google.aip.dev/141

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
-	CRYPTO_ALGORITHM_AES128       = 1;
-	CRYPTO_ALGORITHM_AES192       = 2;
-	CRYPTO_ALGORITHM_AES256       = 3;
-	CRYPTO_ALGORITHM_AES128GCM128 = 4;
-	CRYPTO_ALGORITHM_AES256GCM128 = 5;
-	CRYPTO_ALGORITHM_AES128GMAC   = 6;
-	CRYPTO_ALGORITHM_AES256GMAC   = 7;
-}
-/* Cryptographic algorithm for authentication */
+
```
artek-koltun, line 14: we do not generate Jave bindings at the moment, but we could place Java related options here in advance, like
```proto
option java_package = "opi_api.security.v1";
option java_multiple_files = true;
option java_outer_classname = "IpSec";
```

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
+  optional uint32 mtu = 2;
+}
+
+// IKE SA soft lifetime. Two lifetime values can be configured, either rekey
+// time of the IKE SA or reauth time of the IKE SA. When the rekey lifetime
+// expires, a rekey of the IKE SA starts. When reauth lifetime expires, an IKE
+// SA re-authentication starts.
+message IkeSaLifetimeSoft {
+  // Time in seconds between each IKE SA rekey. The value of 0 means infinite.
+  uint32 rekey_time = 1;
```
artek-koltun, line 224: Would it make sense to use `google.protobuf.Timestamp` for time related vars?

security/ipsec.proto
```diff
@@ -5,477 +5,723 @@
...
-	repeated IntegAlgorithm integ_alg = 2;
-	repeated PRFunction prf = 3;
-	repeated DHGroups dhgroups = 4;
+message CreateIkePeerResponse {}
+
+message UpdateIkePeerRequest {
+  // Name identifying the IKE peer to update
+  string name = 1;
+  // Updated IKE peer specification
+  IkePeer peer = 2;
```
artek-koltun, line 619: Updates requests should contain `update_mask` field https://google.aip.dev/134

artek-koltun (commented):
Thank you, @stevedoyle for the great patch!
We are trying to stick to [AIP](https://google.aip.dev/general) proto design rules and in some places here, we are not consistent.

artek-koltun (commented):
You can use this cmd to run api linter locally to detect the places where api is inconsistent:
```bash
curl -kL https://github.com/googleapis/googleapis/archive/master.tar.gz | tar --strip=1 -zxvf - googleapis-master/google/api
docker run -it --user=$(id -u):$(id -g) --rm --entrypoint=sh -v "${PWD}/../network/opinetcommon":/common -v "${PWD}":/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.56.1 -c "api-linter -I /common /out/*.proto --output-format github --set-exit-status"
```

security/Makefile
sandersms: Since we had a v1 for the security api, we discussed in the api meeting and felt that having a v2alpha1 with the major change would be good to address this move in the security api since it changed things to the CRUD model.  @artek-koltun should we use something else - v1alpha2? instead?

artek-koltun: I am considering if we need to update the version here at all. In other domains, when we had a breaking change, we did not bump the versions. This was because the API was in a state of frequent disruption, which could skyrocket the version numbers if we were to update them each time. In addition, I think, there are no users of this old API, and no one is harmed if we keep the old v1alpha1

sandersms: Good point.  @stevedoyle lets just move the version to v1aplha1 since currently the domains have not been tracking that change level.  This will make it consistent with the other domains until we setup an enforcement for versioning on the APIs.

stevedoyle: Perfect. I've updated the version to v1alpha1

security/Makefile
stevedoyle: I'll raise a separate PR for making the makefiles consistent across folders.

sandersms (approved):
Updates look good - approved

security/Makefile
```diff
@@ -2,27 +2,34 @@
...
 
 	rm -rf "${PWD}"/google
 
 apilint:
 	curl -kL https://github.com/googleapis/googleapis/archive/master.tar.gz | tar --strip=1 -zxvf - googleapis-master/google/api
-	docker run --user=$(id -u):$(id -g) --rm --entrypoint=sh -v "${PWD}":/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.56.1 -c "api-linter -I /common /out/*.proto --output-format github --disable-rule=core::0192 --disable-rule=core::0191 --disable-rule=core::0140 --disable-rule=core::0123 --disable-rule=core::0142 --disable-rule=core::0126 --disable-rule=core::0127 --disable-rule=core::0141 --disable-rule=core::0203 --disable-rule=core::0216 --set-exit-status"
+	docker run --user=$(id -u):$(id -g) --rm --entrypoint=sh -v "${PWD}/../network/opinetcommon":/common -v "${PWD}":/out -w /out ghcr.io/docker-multiarch/google-api-linter:1.56.1 -c "api-linter -I /common /out/*.proto --output-format github --disable-rule=core::0192 --disable-rule=core::0191 --disable-rule=core::0140 --disable-rule=core::0123 --disable-rule=core::0142 --disable-rule=core::0126 --disable-rule=core::0127 --disable-rule=core::0141 --disable-rule=core::0203 --disable-rule=core::0216 --set-exit-status"
```
artek-koltun, line 28: I suppose you fixed many linter issues, so could you please remove as many `--disable-rule` as possible

security/ipsec.proto
artek-koltun: Sorry, I had a typo in the message:
> use of UNsigned types is discouraged: https://google.aip.dev/141

I see the lint type check is disabled for soem uint32/uint64 members. Use of unsigned types is discouraged since some languages do not support unsigned types, like python, java.

Pls confirm that we cannot use signed types instead

artek-koltun (commented):
Looks good, some small notes

security/Makefile
stevedoyle: Fixed. All --disable-rule rules have now been removed.

security/ipsec.proto
stevedoyle: I had reviewed the AIP#141 rule and removed all use of unsigned except for:

- IPsec SPI which is a 32-bit value with a minimum value of 0.
- Sequence numbers. These also have a minimum value of 0 / can't be negative.

I added linter disable for these to highlight that these were deliberate. Alternatives that I had considered:

1. Make these signed types in the API. Calling / implementing code will then need to handle the correct conversion to/from negative values as appropriate. I'm not a fan of this approach as it contradicts the valid range of these fields from the IPsec RFC descriptions and makes the API a little cumbersome.
2. Use a larger sized field, (int64 for SPI and int128 for sequence number) to allow for the API to support the required value range for these fields. The downside is that the fields are now larger than needed.

@artek-koltun Any preference or guidance on the best approach to take here?

artek-koltun: Unfortunately, I don't know an ideal solution for that issue. I wanted to make sure that we considered the option to use the signed types for these cases, and they do not fit

security/Makefile
artek-koltun: Great! Thanks

security/Makefile
artek-koltun: @stevedoyle created an issue to track: https://github.com/opiproject/opi-api/issues/457

Issue #444, opened 2024-02-23
Include https://github.com/googleapis/api-linter in protobuf checking.


sandersms:
The [google-api-linter](https://github.com/googleapis/api-linter) is not run as part of the buf lint.  Buf lint does not include all of the rules in the google linter and currently won't be able to be extended.  We can utilize the linter in the make process to provide more comprehensive checking until buf extends additional checking.

PR #454, opened 2024-06-26
ci: add apilinter

artek-koltun:
Close https://github.com/opiproject/opi-api/issues/444

While there is no google api linter integration with buf, we can run it in our make file separately.

https://github.com/opiproject/opi-api/issues/455


sandersms (approved):
Looks like a good addition for api linting that is missing from buf currently.

Issue #457, opened 2024-07-01
Align Makefiles across folders


artek-koltun:
https://github.com/opiproject/opi-api/pull/453/files#diff-6ea2995538e31968d8342b9ddce98f3e4104f9e6a08d4e62acdedc9535173137

PR #458, opened 2024-07-01
Align Makefiles across folders

stevedoyle:
Reformatting Makefiles across folders to:
- Use an APIVER variable for the API version
- Break long commands into multiple lines to reduce the line lengths
- Have all Makefiles have the same format
- Change inventory/Makefile to a unix file format

Fixes: #457


artek-koltun (approved):
Thank you @stevedoyle

PR #461, opened 2024-07-19
feat(ospf): adding support for OSPF protocol

KeithWiles:
A new feature to support for OSPFv2 protocol using CRUD type APIs. The current APIs support the basic set of APIs for OSPF. These currently do not support all possible OSPF configurations.

Signed-off-by: Keith Wiles <keith.wiles@intel.com>
Signed-off-by: Chunduri, Uma S <uma.s.chunduri@intel.com>
Signed-off-by: Birkner, Matthew <matthew.birkner@intel.com>


artek-koltun (commented):
Hej, thank you for the contribution. Since you are reworking the cloud, could I ask you to remove disabled api linter checks [here](https://github.com/opiproject/opi-api/blob/9920eb170a0728986e197160ba6301dff8fa6cf8/network/cloud/Makefile#L32) so we are confident that the new code is [AIP](https://google.aip.dev/) compliant

KeithWiles:
Hi @artek-koltun  I did not update the Makefile, but I can remove the disable and make locally or did you want me to include that change in my PR?

artek-koltun:
> Hi @artek-koltun I did not update the Makefile, but I can remove the disable and make locally or did you want me to include that change in my PR?

I would prefer to have the cloud Makefile updated as a part of this PR

artek-koltun (approved):
After discussion, we decided to resolve the linter issues in a follow-up PR

Issue #414, opened 2023-12-07
Keep bufbuild-go-gen image under opiproject organization repo


artek-koltun:
see https://github.com/opiproject/opi-api/pull/404#discussion_r1411719342

sandersms:
@glimchb Do we have a opiproject repo where we can place any tools that we have to create for version control and updates?

glimchb:
@sandersms keep it here in this repo under https://github.com/opiproject/opi-api like we do with all images

for example
```
$ docker pull ghcr.io/opiproject/opi-api:v0.1.1
```

we can write a separate action to build and push it on changes in the docker file

like we do in other repos...

sandersms:
The bufbuild-go-gen is moved into the ghcr.io.  It can be used by:

`$ docker pull ghcr.io/sandersms/bufbuild-go-gen:latest`

PR #469, opened 2025-02-27
refactor: transition to buf v2 for api generation

sandersms:
Transition API generation to buf v2 support.  This requires changes to

1) Workflows
2) Structure for the generated files.  The new generated files will move to the ./APIVER/gen folder at the top level
3) Top level makefile used to generate the api files
4) New buf CI which will perform the checking of the api protobuf files
5) Formatting changes that check the protobuf files against industry standards (some files have been run through "buf format")

Once all of the protobuf files have been properly formatted and checked using "buf format", the CI check for format can be enabled.

The existing generated file locations in each of the folders (inventory, network, storage, security) will be maintained in order to keep the bridge prototypes in the repositories working.  These will be transitioned over time and eventually removed as the bridge prototypes move to the new structure.


Makefile
```diff
@@ -0,0 +1,21 @@
...
+bufgen:
+	docker run --rm \
+		-v "${PWD}":/base \
+		-v "${PWD}":/out \
+		-w /out \
+		ghcr.io/sandersms/bufbuild-go-gen:latest generate --template /base/buf.gen.yaml -o ${APIVER}
+
+bufformat:
+	# format the protobuf using buf industry standard
+	buf format --diff -w
```
artek-koltun, line 21: if we have formatting in the makefile:
Would it make sense to force formatting in the github action?
Should we add `bufformat` to `all`?
Can we also used containerized version here: `bufbuild/buf:1.50.0`?

Makefile
```diff
@@ -0,0 +1,21 @@
...
+
+buflint:
+	docker run --rm -v "${PWD}":/out -w /out bufbuild/buf:1.50.0 lint
+
+bufgen:
+	docker run --rm \
+		-v "${PWD}":/base \
+		-v "${PWD}":/out \
+		-w /out \
+		ghcr.io/sandersms/bufbuild-go-gen:latest generate --template /base/buf.gen.yaml -o ${APIVER}
```
artek-koltun, line 17: just a reminder: https://github.com/opiproject/opi-api/issues/414

Pls also re-check if we have go-gen in `bufbuild/buf:1.50.0`

README.md
```diff
@@ -1,9 +1,6 @@
 # OPI API and Behavioral Model Group
 
 [![Linters](https://github.com/opiproject/opi-api/actions/workflows/linters.yml/badge.svg)](https://github.com/opiproject/opi-api/actions/workflows/linters.yml)
-[![OPI Storage API CI](https://github.com/opiproject/opi-api/actions/workflows/storage.yml/badge.svg)](https://github.com/opiproject/opi-api/actions/workflows/storage.yml)
```
artek-koltun, line 4: should we add a badge for [buf.yml](https://github.com/opiproject/opi-api/pull/469/files#diff-508178dd28549c8a03868db7b10cbc0af552c6371f0413ab878e32b1de325b46)?

buf.yaml
```diff
@@ -0,0 +1,20 @@
...
+    - STANDARD
+    - COMMENTS
+  except:
+    - PACKAGE_DIRECTORY_MATCH
+    # Don't check standard name as that causes google aip issues
+    - RPC_RESPONSE_STANDARD_NAME
+    # Allow same name used in request/response type for multiple RPCs
+    - RPC_REQUEST_RESPONSE_UNIQUE
+    - ENUM_ZERO_VALUE_SUFFIX
+    # disabled comment checking - todo: enable and resolve these errors
```
artek-koltun, line 14: do we have an issue for that?

Makefile
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (C) 2022 Intel Corporation
+# Copyright (c) 2022 - 2025 Dell Inc, or its subsidiaries.
+
+APIVER = v1
+
+all: buflint bufgen
+
+buflint:
+	docker run --rm -v "${PWD}":/out -w /out bufbuild/buf:1.50.0 lint
```
artek-koltun, line 10: We had google api linter as `apilint` target. Would it make sense to add it here until we have aep plugin running as a replacement?

sandersms:
Moving this forward to allow for the transition to the Buf V2 and the Buf CI processing

Makefile
sandersms: All of the pieces are not in the buf:1.50.0 that can be used.

Makefile
sandersms: I put it in the makefile to allow for someone to do the format step and make the appropriate corrections.  The format command will force the write of the changes with the -w.  You can just do a --diff to see what the changes are.  Currently there are files in the .proto format that have not been formatted completely proper, so that still needs to be done as part of the V2 updates.

For the buf format --diff -w I will look at the buf:1.50.0 version also.  I have it locally, but it may also be valuable in the container view.  Let me investigate that separately.

buf.yaml
sandersms: Not yet - that is on my list of things to add - additional issues for the Buf V2 transition.

PR #485, opened 2025-12-17
Add lifecycle management APIs

bn222:
Add lifecycle management APIs

- LifeCycleService: initialization and configuration
- DeviceService: Device enumeration and VF management
- HeartbeatService: Health monitoring and status checks

These APIs provide standardized interfaces for:
- Platform initialization (Init RPC)
- Device discovery and configuration (GetDevices, SetNumVfs RPCs)
- Health monitoring (Ping RPC)


buf.gen.yaml
```diff
@@ -1,13 +1,13 @@
 version: v2
 plugins:
-  - remote: buf.build/protocolbuffers/go:v1.36.3
-    out: gen/go
+  - remote: buf.build/protocolbuffers/go:v1.36.11
+    out: v1/gen/go
```
sandersms: The top level makefile has the APIVER variable that allows for the version of the API to change over time.  Adding the v1 here breaks the generation that is provided in the makefile and forces a change each time that the API version has to change.  

This needs to be reverted back to maintain the ability for the makefile to generate the files in the proper directory and eliminate the need to change the file when the API Changes.

buf.gen.yaml
```diff
@@ -1,13 +1,13 @@
...
     opt: paths=source_relative
   - remote: buf.build/grpc/go:v1.5.1
-    out: gen/go
+    out: v1/gen/go
```
sandersms: This breaks the makefile ability to generate the files to the proper directory location. (see above).

buf.gen.yaml
```diff
@@ -1,13 +1,13 @@
...
     opt: paths=source_relative
   - remote: buf.build/grpc-ecosystem/gateway:v2.26.0
-    out: ./gen/go
+    out: v1/gen/go
```
sandersms: This breaks the makefile ability to generate the files to the proper directory location (see above).

v1/gen/python/lifecycle/lifecycle_pb2_grpc.py
sandersms, line 1: Why is this being loaded into the v1alpha1 folder?  It should be in the top level lifecycle folder.

lifecycle/lifecycle.proto
sandersms, line 1: Why is the .proto file in the v1alpha1 folder?  It should be in the lifecycle folder.

v1/gen/go/inventory/go.sum
sandersms, line 1: Why is this file needed in the API repository?  Only the go.mod file is necessary for importing into the projects.  This is the same for all of the other folders under the v1 folder version of the API

sandersms (changes requested):

sandersms:
The last checkin for the refactor should be removed from this PR and submitted separately as it breaks many things in the API building from the makefile.

bn222:
@sandersms ready for another round

bn222:
@sandersms let me know if more changes are needed.

wizhaoredhat:
LGTM @sandersms / @pdp2shirts could you help merge it?
We have a lot of code to contribute (1000s of lines of go code from downstream). And we need a way to move fast. WDYT?

v1/gen/python/lifecycle/lifecycle_pb2_grpc.py
sandersms: Looks Good

sandersms (approved):
These changes look good now.

PR #486, opened 2025-12-18
fix: correct markdown table formatting

bn222:
Make the linter pass

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run tests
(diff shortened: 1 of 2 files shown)


bn222:
@sandersms can you review this one? It's a very small change that fixes head.

wizhaoredhat:
LGTM @sandersms / @pdp2shirts could you help merge it?
We have a lot of code to contribute (1000s of lines of go code from downstream). And we need a way to move fast. WDYT?

sandersms (approved):
This is fine - perhaps in the future we can re-enable the checks and resolve the issues in the markdown files.

bn222:
That alternative (re-enable + resolve) leads to a very hard to maintain md imo, and for that reason I've chosen the path to disable the check. Leaving this comment here for the future so that we consider it as input to create an issue to tackle this.

